PR #15, opened 2021-01-19
TILA-247 reservation unit search enhancements

kieferbonk:


src/application/page1/ReservationUnitModal.tsx
```diff
@@ -125,63 +132,156 @@ const Text = styled.span`
...
+          options={reservationUnitTypeOptions}
+          label={t('ReservationUnitModal.searchReservationUnitTypeLabel')}
+          onChange={(selection: OptionType): void => {
+            setReservationUnitType(selection);
+          }}
+          defaultValue={emptyOption}
+        />
+      </Filters>
+      <ButtonContainer>
+        <SearchButton onClick={() => searchResults()}>
```
siren, line 262: Looks like this button now tries to submit the form behind the modal tool, need to add call to "e.preventDefault()' ?

siren (commented):
There's something strange happening with the Spinner, sometimes I see the following error when clicking 'Hae'

```
LoadingSpinner-7786d9a3.js:1 Uncaught TypeError: Cannot read property 'removeChild' of null
    at LoadingSpinner-7786d9a3.js:1
```

src/application/page1/ReservationUnitModal.tsx
```diff
@@ -125,63 +132,156 @@ const Text = styled.span`
...
 
-  const results = useAsync(async () => {
-    if (q === null) {
-      return [];
-    }
-    return getReservationUnits({ search: q });
-  }, [q]);
+  const searchResults = async () => {
+    setSearching(true);
+    const searchCriteria = {
```
siren, line 208: What about the ApplicationPeriod filtering?

kieferbonk: Forgot about that - added.

src/application/page1/ReservationUnitModal.tsx
kieferbonk: Yes, good catch. Fixed.

kieferbonk:
I was unable to reproduce the spinner problem. Updated hds to new version, that adressed some spinner bug.

src/application/page1/ReservationUnitModal.tsx
siren: Not super important right now, but the api also supports filtering with parameter application_period , for example application_period=1

siren (approved):

siren (approved):

PR #17, opened 2021-01-21
TILA-249: time selection summary

kieferbonk:

PR #18, opened 2021-01-21
TILA-254 display events inside accordions

siren:
- also changed getting of reservation unit data so that only the required data is retrieved
- handling of null values


kieferbonk:
Looks good, but maybe we should wait for https://github.com/City-of-Helsinki/tilavarauspalvelu-ui/pull/17 to clear before merging this.

kieferbonk (approved):

PR #23, opened 2021-01-31
TILA-336 reservation unit filtering

siren:


src/common/util.ts
```diff
@@ -1,12 +1,30 @@
...
   const now = new Date();
   return isAfter(now, parseISO(startDate)) && isBefore(now, parseISO(endDate));
 };
 
+const isPast = (endDate: string): boolean => {
+  const now = new Date();
+  return isAfter(now, parseISO(endDate));
+};
+
+export const uiState = (
```
kieferbonk, line 14: This could use a bit more descriptive name.

src/search/SearchForm.tsx
```diff
@@ -1,16 +1,21 @@
...
+import { mapOptions, OptionType, getSelectedOption } from '../common/util';
 
-interface Props {
-  // only text search is now implemented!
-  onSearch: (text: string) => void;
-}
-interface OptionType {
-  label: string;
-}
+export type Criteria = {
```
kieferbonk, line 10: Maybe not needed to be exported, or if so, belongs to to the type definition file?

kieferbonk (approved):

src/search/SearchForm.tsx
siren: fixed

src/common/util.ts
siren: true, renamed

PR #25, opened 2021-02-04
TILA-130 e2e

siren:
contains:
- e2e test template
- means to run e2e tests against local setup


kieferbonk (changes requested):
The test running looking good 👍 
I think the build is failing because of config changes and/or dependencies needed for post-build operations, that were moved to dev-deps.

kieferbonk (approved):

PR #31, opened 2021-02-10
changes to the .env.production file, .env.developer file, Dockerfile

jaana-ibm:
Changes that have been made:

- changed hard-coded values ​​for variables
- changed permissions to the /src folder
- added RUN command to change $ {REACT_APP_SPACE RESERVATION_API_URL} value


siren (commented):
It's currently difficult to give proper feedback on this because I still have not seen how the devops part is configured or built.

Still I think there's something logically incorrect with this approach, the .env.development is supposed to be used when developing the app locally. Why is it changed?

I think that the Sentry DSN should also be set to a reasonable value at some stage of the process. One option would be to commit the configuration to the git repository or do you wish to set it up with environment variable?

siren:
I think this is not needed anymore -> closing.

PR #34, opened 2021-02-15
TILA-240

siren:
Add more search filters to reservation unit search


src/search/SearchForm.tsx
```diff
@@ -100,6 +128,12 @@ const SearchForm = ({ onSearch, formValues }: Props): JSX.Element | null => {
     Object.keys(formValues).forEach((p) => setValue(p, formValues[p]));
   }, [formValues, setValue]);
 
+  useEffect(() => {
+    if (showMoreFilters) {
+      document.getElementById('participantCountFilter-toggle-button')?.focus();
```
kieferbonk, line 133: No biggie, but could this be resolved with a `ref`?

kieferbonk (approved):

src/search/SearchForm.tsx
siren: yeah, I tried to figure that out, but was not able to because select does not support ref.

PR #38, opened 2021-02-26
Refactoring & cleanup

siren:
- selected reservation units are now persisted to localstorage so that they survive during auth
- completed todos in application save
- finished applicationPeriod -> applicationRound


src/i18n/messages.ts
```diff
@@ -129,11 +129,30 @@ const translations: Translations = {
       },
     },
     preview: {
+      applicantType: {
+        community: [
+          'Rekitesteröitymätön yhdistys',
```
kieferbonk, line 134: Typo

kieferbonk (approved):
👍

PR #39, opened 2021-03-01
TILA-392

siren:
- Adds a map & map marker to show the location of the reservation unit in the reservation unit page


src/component/MapMarkerIcon.tsx
```diff
@@ -0,0 +1,48 @@
...
+          d="M15.9981139,0 C20.1340105,0 24.2713477,1.53717334 27.4262114,4.61293808 C33.7359387,10.7630495 33.3095279,20.347694 27.4262114,26.8863529 L26.174361,28.2877277 C20.9310906,34.2007901 18.0099826,38.091836 15.9981139,42 C13.8430107,37.791208 10.613237,33.6022689 4.57433813,26.8863529 C-1.30897841,20.347694 -1.73682979,10.7630495 4.57433813,4.61293808 C7.7292018,1.53717334 11.8650984,0 15.9981139,0 Z M16.0657464,10.8717672 C13.2952057,10.8717672 11.0444238,13.0895872 11.0444238,15.8145918 C11.0444238,18.5373718 13.2952057,20.7551919 16.0657464,20.7551919 C18.8340272,20.7551919 21.0848091,18.5373718 21.0848091,15.8145918 C21.0848091,13.0895872 18.8340272,10.8717672 16.0657464,10.8717672 Z"
+          id="selectable"
+          fill="#0000BF"
+        />
+      </g>
+    </g>
+  </svg>
+);
+
+export default MapMarkerIcon;
```
kieferbonk, line 48: If you're ok with static content, a more straight forward way would be to save the svg as is, and then webpack can import it like so: `import { ReactComponent as Icon } from "../../images/icon.svg";`

kieferbonk (approved):

src/component/MapMarkerIcon.tsx
siren: cool, thanks!

PR #43, opened 2021-03-11
Retrieve and use api specific token when accessing tilavaraus-api

siren:


src/common/auth/LoggingIn.tsx
```diff
@@ -0,0 +1,16 @@
...
+  return (
+    <BrowserRouter>
+      <PageWrapper>
+        <LoadingSpinner />
+      </PageWrapper>
+    </BrowserRouter>
+  );
+};
+
+export default UpdateToken;
```
kieferbonk, line 16: I think you forgot to rename the component.

src/common/auth/axiosClient.ts
```diff
@@ -0,0 +1,86 @@
...
+    headers: {
+      Authorization: `Bearer ${accessToken}`,
+      'Content-Type': 'application/x-www-form-urlencoded',
+    },
+  });
+
+  const { data } = response;
+
+  const apiAccessToken = data[apiScope];
+  setApiAccessToken(apiAccessToken, 'https://api.hel.fi/auth/tilavarausapidev');
```
kieferbonk, line 50: You probably want to configure the url per environment.

kieferbonk (changes requested):

src/common/auth/LoggingIn.tsx
siren: fixed

src/common/auth/axiosClient.ts
siren: fixed

kieferbonk (approved):

PR #89, opened 2021-08-20
TILA-671 space creation modal, initial impl

siren:
- creation now works, there are still some values hardwired because  api lacks properties or requires something that the ui does not provide


admin-ui/src/component/Unit/Modal.tsx
```diff
@@ -0,0 +1,53 @@
...
+export default Modal;
+
+export const useModal = (
+  initiallyOpen = false
+): [boolean, () => void, () => void] => {
+  const [open, setOpen] = useState(initiallyOpen);
+  const openDialog = () => setOpen(true);
+  const closeDialog = () => setOpen(false);
+  return [open, openDialog, closeDialog];
+};
```
kieferbonk, line 53: This looks like it could be used as a more general component as opposed to the connection to `unit`. So a naming/placement change would to it good,  I think.

kieferbonk (approved):

PR #128, opened 2021-11-29
bugfixing

siren:
fixes following issues:
- TILA.-932 (link to reservation unit not working)
- TILA-931 (fix toas message)
- TILA-930 (display toast when creatinh new reservation unit)

### DIFF
--- a/admin-ui/src/component/Unit/SpacesTable.tsx
+++ b/admin-ui/src/component/Unit/SpacesTable.tsx
@@ -252,6 +252,7 @@ const SpacesTable = ({
   return (
     <Wrapper>
       <DataTable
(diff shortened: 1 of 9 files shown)


admin-ui/src/component/Unit/SpacesTable.tsx
```diff
@@ -252,6 +252,7 @@ const SpacesTable = ({
   return (
     <Wrapper>
       <DataTable
+        key={spaces.length}
```
kieferbonk, line 255: Doesn't this prevent rendering after sorting by column, for example?

kieferbonk (approved):

PR #143, opened 2021-12-13
TILA-1039 reservation unit prices

kieferbonk:


ui/mocks/handlers/singleSearch.ts
```diff
@@ -59,6 +64,10 @@ export const reservationUnitSearchHandlers = [
               nameFi: "Hertsin nuorisotalon sali",
               nameEn: "Hertsin nuorisotalon sali",
               nameSv: "Hertsin nuorisotalon sali",
+              lowestPrice: 0.0,
+              highestPrice: 20,
+              priceUnit:
+                "PER_HALF_DAY" as ReservationUnitsReservationUnitPriceUnitChoices,
```
siren, line 70: We have a generated typescript enum (in this case ReservationUnitsReservationUnitPriceUnitChoices.PerHalfDay). 

By using the enum values we would catch breaking api changes better, no?

siren: never mind, there's no difference between the two styles

siren (approved):

kieferbonk:
Squashed into main.

PR #158, opened 2022-01-10
TILA-1038

siren:
- make sure api requests are authenticated from the start
- refactored contexts


admin-ui/src/App.tsx
```diff
@@ -1,13 +1,12 @@
-import React, { useState } from "react";
+import React from "react";
 import { BrowserRouter, Route, Switch } from "react-router-dom";
 // eslint-disable-next-line import/no-unresolved
```
kieferbonk, line 3: This can be probably cleaned up.

kieferbonk (approved):
Nice improvement!

PR #166, opened 2022-01-20
Fix for duration api changes

siren:
- also make error notification dismissible


kieferbonk (changes requested):
The schema of reservation unit's query and mutation input/payload have changed - shouldn't they also be updated?

siren:
yeah, I quess ->  regenerated the types

kieferbonk (approved):

PR #206, opened 2022-03-01
TILA-1223

siren:
reorganizing urls, navigation, directory structure


kieferbonk (approved):
Looks good, although linter caught a bunch of errors -- maybe the project linting config should be revised.

PR #278, opened 2022-04-11
TILA-1478, TILA-1479

siren:
- truncate applicant name
- use applicant name from form data (in listing and also on details page)

### DIFF
--- a/admin-ui/src/component/applications/util.ts
+++ b/admin-ui/src/component/applications/util.ts
@@ -9,12 +9,17 @@ import {
 } from "../../common/types";
 import { formatDuration } from "../../common/util";
 
-export const applicantName = (
-  application: Application | null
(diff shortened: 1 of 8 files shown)


admin-ui/src/component/applications/util.ts
```diff
@@ -9,12 +9,17 @@ import {
...
 
-export const applicantName = (
-  application: Application | null
-): string | null | undefined =>
-  application?.applicantType === "individual"
-    ? application?.applicantName
-    : application?.organisation?.name;
+export const applicantName = (app: Application | null): string => {
+  if (app === null) {
+    return "null";
```
kieferbonk, line 14: Return `null`-string?

kieferbonk (approved):

PR #332, opened 2022-06-10
TILA-1712, TILA-1556

siren:
- refactored graphql listings code to be more reusable (paging, search tags, layout)
- applications view now uses gql (missing application events -> splitting to another issue)
- units view has new layout and also uses gql now


kieferbonk (approved):
A bit too large pr to get grips on, but 👍

PR #363, opened 2022-08-15
TILA-1278 update auth-flow

kieferbonk:


siren (approved):
related readme &  config example could be updated with new values

PR #480, opened 2022-12-13
Tila 2046 - ReservationUnits calendar event title

wuahi:
- Updates events title when the reservation is from another reservationUnit from what is being viewed
 - Fixes a bug with the sorting of reservation units when title is undefined
 - Fixes a bug with begin date in comparing the events that should be in view

### DIFF
--- a/admin-ui/src/component/ReservationUnits/ReservationUnitEditor/SortedSelect.tsx
+++ b/admin-ui/src/component/ReservationUnits/ReservationUnitEditor/SortedSelect.tsx
@@ -2,6 +2,7 @@ import React from "react";
 import { useTranslation } from "react-i18next";
 import memoize from "lodash/memoize";
 import { Select, SelectProps } from "hds-react";
+import { sortByName } from "../../../common/util";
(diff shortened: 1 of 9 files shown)


admin-ui/src/component/ReservationUnits/ReservationUnitEditor/SortedSelect.tsx
```diff
@@ -12,7 +12,11 @@ function SortedSelect<T>(
     const opts = [...originalOptions];
     if (props.sort) {
       opts.sort((a, b) =>
-        a.label.toLowerCase().localeCompare(b.label.toLowerCase())
+        a.label && b.label
```
kieferbonk, line 15: Maybe the sorting logic could be moved to a helper function to minimize nested ternarys in code - applies also to the one in [admin-ui/src/component/my-units/UnitCalendar.tsx]

kieferbonk (changes requested):

kieferbonk (approved):

PR #482, opened 2022-12-15
Tila 2129 - Base for customer emails

wuahi:
- Adds mjml library for creating html templates
 - Adds simple nodejs script for easy .mjml -> .html generation
 - Adds first html template: ReservationConfirmation


vergama (changes requested):
One more thing, there should be line for unit_name (Toimipaikka) between reservation unit name (Varausyksikkö) and the location/address (which has been imported from TPRek at the unit level in the model):
<mj-text>{{reservation_unit}}</mj-text>
<mj-text>{{unit_name}}</mj-text>
<mj-text>{{unit_location}}</mj-text>
=>
Ryhmätila Dekkarikirjasto
Kallion Kirjasto
Viides linja 11, 01500 Helsinki

vergama (changes requested):
Sorry, I overlooked `{{name}}` in this context, we can do away with that since it's the name of the _reservation_ itself (optionally filled in by reservee for own purposes beside a longer description), as opposed to the name of the _reservation unit_ (which the Figma spec says should be in boldface). Confusing, eh - let the `{{name}}` be available in the [variables supplied by core](https://github.com/City-of-Helsinki/tilavarauspalvelu-core/blob/main/email_notification/sender/email_notification_builder.py#L98) but the email templates as now defined no longer make use of it for identification, just the reservation number.
The variable naming might be somewhat different if it had been made from scratch now...

So finally:
<mj-text><b>{{name}}</b></mj-text>
<mj-text>{{reservation_unit}}</mj-text>
<mj-text>{{unit_name}}</mj-text>
<mj-text>{{unit_location}}</mj-text>
=>
<mj-text><b>{{reservation_unit}}</b></mj-text>
<mj-text>{{unit_name}}</mj-text>
<mj-text>{{unit_location}}</mj-text>

vergama (approved):

PR #483, opened 2022-12-15
TILA-2048 inline pricing terms in reservation funnel

kieferbonk:


common/types/gql-types.ts
```diff
@@ -1866,7 +1866,7 @@ export type QueryReservationsArgs = {
   offset?: InputMaybe<Scalars["Int"]>;
   onlyWithPermission?: InputMaybe<Scalars["Boolean"]>;
   orderBy?: InputMaybe<Scalars["String"]>;
-  paymentStatus?: InputMaybe<Array<InputMaybe<Scalars["String"]>>>;
+  orderStatus?: InputMaybe<Array<InputMaybe<Scalars["String"]>>>;
```
wuahi, line 1869: wasn't this already fixed O_O

kieferbonk: Suppose they're still at PR level. `orderStatus` is the current one, right?

wuahi: Yep. strange. the pr has gone through like a week ago.

wuahi (approved):

PR #524, opened 2023-01-25
TILA-2236 toggle controls on time selection

kieferbonk:


ui/components/calendar/ReservationCalendarControls.tsx
```diff
@@ -493,7 +523,12 @@ const ReservationCalendarControls = <T extends Record<string, unknown>>({
             )}
           </TogglerLabel>
           <ToggleButton
-            onClick={() => setAreControlsVisible(!areControlsVisible)}
+            onClick={() => {
+              setAreControlsVisible(!areControlsVisible);
```
wuahi, line 527: Why two different states? are and should?

ui/pages/reservation-unit/[id].tsx
```diff
@@ -569,6 +569,14 @@ const ReservationUnit = ({
     );
   }, [reservationUnit.canApplyFreeOfCharge, reservationUnit.pricings]);
 
+  const isTouchDevice = useMemo(
```
Comment (wuahi) [inline ui/pages/reservation-unit/[id].tsx:572]: Could this be a utility function? then there would possibly be no need for shouldCalendarControlsBeVisible? could just use the same function?

wuahi (approved):

PR #534, opened 2023-02-09
TILA-2270 reflect opening times api change; support multi day events

kieferbonk:


common/src/calendar/util.ts
```diff
@@ -109,17 +109,13 @@ const areOpeningTimesAvailable = (
...
-    return (
-      toApiDate(slotDate) === startDate?.toString() &&
-      slotDate < endDateTime &&
-      startDateTime.getDay() === slotDate.getDay() &&
-      startDateTime.getHours() <= slotDate.getHours() &&
-      endDateTime.getHours() >= slotDate.getHours()
-    );
+    const { startTime, endTime } = oh;
+    const startDate = new Date(startTime as string);
+    const endDate = new Date(endTime as string);
```
joonatank, line 116: Type cast looks dangerous.
new Date(undefined) => will cause a run time error
explicit null check and return false early so it doesn't need 'as'

common/src/calendar/__tests__/calendar.test.ts
```diff
@@ -658,7 +661,7 @@ describe("isReservationUnitReservable", () => {
         metadataSet: {
           supportedFields: ["name"],
         },
-      } as ReservationUnitByPkType)
+      } as unknown as ReservationUnitByPkType)
```
joonatank, line 664: Same as the other PR. Not fan of type casting, common props with spread is better.

common/src/calendar/util.ts
```diff
@@ -482,19 +472,18 @@ export const getAvailableTimes = (
...
 
   const times: string[] = intervals.map((val) => {
-    const [startHours, startMinutes] = val.split(":").map(Number);
-
-    const start = new Date(date);
-    start.setHours(startHours, startMinutes);
+    const [startHours, startMinutes] = val.split(":");
 
-    return toUIDate(start, "HH:mm");
+    return `${startHours}:${startMinutes}`;
```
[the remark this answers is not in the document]
kieferbonk: That's to get rid of the seconds.

common/src/calendar/__tests__/calendar.test.ts
```diff
@@ -781,9 +828,9 @@ describe("isReservationUnitReservable", () => {
     openingHours: {
       openingTimes: [
         {
-          date: new Date().toISOString(),
-          startTime: `${new Date().toISOString()}T04:00:00+00:00`,
-          endTime: `${new Date().toISOString()}T20:00:00+00:00`,
+          date,
+          startTime: `${date}T04:00:00+00:00`,
+          endTime: `${date}T20:00:00+00:00`,
```
joonatank, line 833: How does this work? how did it even work before?
`${new Date().toISOString()}T20:00:00+00:00` ~= '2023-02-23T07:45:05.193ZT20:00:00+00:00'
So it has double T... strings.

kieferbonk: By chance I guess. Corrected.

joonatank (approved):

PR #543, opened 2023-02-24
Tila 2103: Recurring reservation

joonatank:


admin-ui/src/component/my-units/MyUnitRecurringReservation/WeekdaysSelector.tsx
```diff
@@ -0,0 +1,133 @@
...
+    }
+  };
+
+  // NOTE This is not accessible (even with the role="checkbox")
+  return (
+    <div>
+      <Label>{label}</Label>
+      <Wrapper>
+        {weekdays.map((weekday) => (
+          <Day
```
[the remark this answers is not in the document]
joonatank: Replaced it with a button for Day. For accessibility it should be a selection list with checkbox children, but that's another two days of work (selection lists are pita and HDS isn't easy to style).

ui/tsconfig.tsbuildinfo
```diff
@@ -1 +0,0 @@
...62,1],[363,1],[358,1],[360,1],[1166,1],[1169,1],[1164,1],[1168,1],[1170,1],[1167,1],[1165,1],[1160,1],[1163,1],[1162,1],[1161,1],[1525,1],[1528,1],[1527,1],[1526,1],[1414,1],[1325,1],[1327,1],[1322,1],[1321,1],[1334,1],[1326,1],[1330,1],[1331,1],[1332,1],[1320,1],[1319,1],[1323,1],[1324,1],[1328,1],[1333,1],[1329,1],[1352,1],[1359,1],[1368,1],[1353,1],[1373,1],[1362,1],[1365,1],[1360,1],[1364,1],[1366,1],[1363,1],[1361,1],[1367,1],[1358,1],[1355,1],[1357,1],[1351,1],[1354,1],[1356,1],[1376,1],[1382,1],[1375,1],[1380,1],[1379,1],[1381,1],[1377,1],[1374,1],[1378,1],[1372,1],[1371,1],[1370,1],[1369,1],[1416,1],[1417,1],[1426,1],[1418,1],[1425,1],[1424,1],[1419,1],[1421,1],[1422,1],[1423,1],[1420,1],[1404,1],[1427,1],[1403,1],[1384,1],[1383,1],[1335,1],[1349,1],[1350,1],[1336,1],[1343,1],[1342,1],[1337,1],[1345,1],[1340,1],[1341,1],[1338,1],[1339,1],[1344,1],[1348,1],[1314,1],[1346,1],[1315,1],[1316,1],[1317,1],[1318,1],[1347,1],[1397,1],[1398,1],[1396,1],[1388,1],[1387,1],[1402,1],[1399,1],[1395,1],[1389,1],[1394,1],[1392,1],[1391,1],[1390,1],[1393,1],[1385,1],[1400,1],[1386,1],[1401,1],[1241,1],[1243,1],[1283,1],[1277,1],[1245,1],[1247,1],[1244,1],[1248,1],[1246,1],[1264,1],[1278,1],
```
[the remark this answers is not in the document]
joonatank: it's in the root level .gitignore

admin-ui/src/component/my-units/MyUnitRecurringReservation/MyUnitRecurringReservation.tsx
```diff
@@ -0,0 +1,104 @@
...
+            variant="supplementary"
+            iconLeft={<IconAngleLeft />}
+          >
+            {t("common.prev")}
+          </Button>
+        </BasicLink>
+      </PreviousLinkWrapper>
+
+      <Container>
+        {sent != null ? (
```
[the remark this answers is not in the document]
joonatank: Not for null checks since
``` 
(undefined === null) === false // but 
(undefined == null) === true
```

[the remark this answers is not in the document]
kieferbonk: The problem with loose equality operators is that it's pretty close to a typo, if using a strict operator would be viable. In cases like these, where you check against general falsiness, I'd equate to booleans (`!sent`).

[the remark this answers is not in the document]
joonatank: Don't like !sent because.
```
 !"" === true 
!0 === true
```

Linter can't catch these, sometimes "" is valid where null isn't, that's the exact case where you'd do something like 
```
const [var, setVar] = useState<string | null>(null)
```

Linter should be used to catch typos i.e. we should use one of these in eslint
eqeqeq: ["error", "smart"]
eqeqeq: ["error", "always", { null: "never" }]

admin-ui/src/component/my-units/MyUnitRecurringReservation/MyUnitRecurringReservation.tsx
```diff
@@ -0,0 +1,98 @@
...
+        >
+          {t("common.prev")}
+        </Button>
+      </BasicLink>
+    </PreviousLinkWrapper>
+  );
+};
+
+const MyUnitRecurringReservation = ({ unitId }: { unitId: number }) => {
+  const [sent, setSent] = useState<ReservationsMade | null>(null);
```
kieferbonk, line 47: This could use more descriptive name.

admin-ui/src/component/my-units/MyUnitRecurringReservation/RecurringReservationDone.tsx
```diff
@@ -0,0 +1,96 @@
...
+  margin-bottom: 0;
+`;
+
+export type ReservationsMade = Array<{
+  reservationPk?: number;
+  startTime: string;
+  endTime: string;
+  date: Date;
+  error?: string | ErrorType[];
+}>;
```
kieferbonk, line 26: Is there motivation for the type format? Defining it as an object and referring to it as `ReservationsMade[]` would be preferable.

admin-ui/src/component/my-units/MyUnitRecurringReservation/MyUnitRecurringReservation.tsx
```diff
@@ -0,0 +1,99 @@
...
+  return (
+    <>
+      <BackLinkHeader unitId={unitId} />
+      <Container>
+        {reservationMade !== null ? (
+          <RecurringSuccess reservation={reservationMade} />
+        ) : (
+          <>
+            <H1 $legacy>{t("MyUnits.RecurringReservation.pageTitle")}</H1>
+            {reservationUnits != null && reservationUnits.length > 0 ? (
```
kieferbonk, line 64: Using just `reservationUnits?.length > 0` might make it more readable.

admin-ui/src/component/my-units/MyUnitRecurringReservation/RecurringReservationDone.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+type RecurringSuccessProps = {
+  reservation: ReservationsMade;
```
kieferbonk, line 29: Naming and type definition suggests a single reservation item.

admin-ui/src/variables.css
```diff
@@ -31,6 +32,9 @@
   --font-regular: HelsinkiGrotesk, var(--font-default);
   --font-medium: HelsinkiGrotesk-Medium, var(--font-default);
   --font-bold: HelsinkiGrotesk-Bold, var(--font-default);
+  --prose-width: 66ch;
```
kieferbonk, line 35: Is this sensible when using variable-width fonts?

kieferbonk (changes requested):

admin-ui/src/variables.css
joonatank: That's the one suggested by web.dev currently.
It handles scaling fonts with zoom (unlike px), and it handles different font-sizes (unlike rem). 66ch is around 650px with normal 18px latin fonts, but it scales better when the font is larger or smaller.
Max prose size is for legibility. 12px font is not legible at 650px, but 48px font is still legible at 1200px.

So it's the same as rem units but allows us to use 940px (container width) titles instead of clipping them to 650px as with rem units (assuming we use it for titles, currently it's only for text inputs).

For variable width fonts there might be an issue for example Helvetica 0 is thinner than 'm' so a line of only ms will overflow. But the only time this should happen is if we force no line-breaks, otherwise text should wrap automatically. The non contrived example I can think of is limiting a card title to two lines of 20ch and testing it with 0s instead of ms, but it does require forcing manual line-breaks.

admin-ui/src/component/my-units/MyUnitRecurringReservation/MyUnitRecurringReservation.tsx
joonatank: valid JS but TS has a hissy fit about it (both VS code with TS 5.1 and TS playground with 4.9.5). undefined > 0 is not valid.

admin-ui/src/variables.css
kieferbonk: Ok, interesting. Can you do a demo on this next tue?

common/src/reservation-form/ReservationFormField.tsx
```diff
@@ -85,21 +93,42 @@ const StyledCheckbox = styled(Checkbox)`
   }
 `;
 
+const ControlledCheckbox = (props: {
+  field: string;
+  control: Control<FieldValues, any>;
```
kieferbonk, line 98: This should be addressed. Otherwise lgtm.

kieferbonk (approved):

PR #548, opened 2023-02-28
TILA-2350 clear state between navigation

kieferbonk:


ui/pages/reservation-unit/[id].tsx
```diff
@@ -700,7 +719,13 @@ const ReservationUnit = ({
         reservationUnitPks: [reservationUnit.pk],
       };
 
-      setReservation({ begin, end, pk: reservationUnit.pk, price: null });
+      setReservation({
+        begin,
+        end,
+        pk: reservationUnit.pk,
+        price: null,
+        reservationUnitPk: reservationUnit.pk,
```
Comment (joonatank) [inline ui/pages/reservation-unit/[id].tsx:727]: Why is the pk === reservationUnitPk

ui/pages/reservation-unit/[id].tsx
```diff
@@ -402,6 +413,13 @@ const ReservationUnit = ({
     [reservationUnit.canApplyFreeOfCharge]
   );
 
+  useEffect(() => {
+    setFocusDate(new Date());
+    setInitialReservation(null);
+    setIsDialogOpen(false);
+    setPendingReservation(null);
+  }, [dynamicRoute, setPendingReservation, setReservation]);
```
Comment (joonatank) [inline ui/pages/reservation-unit/[id].tsx:421]: Why is this needed? is there a storage that persists through routes? e.g. context
Wouldn't the right solution to be to clear the context on these conditions? or at least wrap it into a named hook so it's clear why and what it does.

ui/pages/reservation-unit/[id].tsx
```diff
@@ -361,14 +363,23 @@ const ReservationUnit = ({
...
   const now = useMemo(() => new Date().toISOString(), []);
 
-  const { reservation, setReservation } = useContext(DataContext);
+  const { reservation: contextReservation, setReservation } =
+    useContext(DataContext);
+
+  const reservation =
+    contextReservation?.reservationUnitPk === reservationUnit.pk
+      ? contextReservation
+      : null;
```
Comment (joonatank) [inline ui/pages/reservation-unit/[id].tsx:382]: I guess this is fine, but this should be more an assert (should never happen) and the context should be cleared when user switches between ReservationUnits. Alternatively the partial reservation (context) should be shown to the user and asked if he wants to replace it with the new one.
Or when user navigates outside the ReservationUnit confirm and clear context.

ui/pages/reservation-unit/[id].tsx
Comment (kieferbonk) [reply ui/pages/reservation-unit/[id].tsx]: Good catch, was some relic that had no effect on anything.

ui/pages/reservation-unit/[id].tsx
Comment (kieferbonk) [reply ui/pages/reservation-unit/[id].tsx]: Turned out that nextjs persists dynamic page state across route changes. What actually helped was the `key` prop to the component/page and the other cleanup was unnecessary.

ui/pages/reservation-unit/[id].tsx
Comment (kieferbonk) [reply ui/pages/reservation-unit/[id].tsx]: Yes, this inspired me to clean up some sedimented logic.

ui/components/reservation/EditStep0.tsx
```diff
@@ -479,14 +471,24 @@ const EditStep0 = ({
...
+              ...initialReservation,
+              price: getReservationUnitPrice({
+                reservationUnit,
+                pricingDate: initialReservation.begin
+                  ? new Date(initialReservation.begin)
+                  : null,
+                minutes: 0,
+                asInt: true,
+              }),
+            } as unknown as ReservationType;
```
joonatank, line 486: I'd remove the type cast. If it's not infered properly then const newReservation: ReservationType should fix it.

joonatank (commented):
Nice. Just the type cast is iffy.

joonatank (approved):

PR #553, opened 2023-03-07
Tila 2290 - Change status tag colours and add new reservationStatus tag next to the unit status tag. 

joonatank:


kieferbonk (approved):

joonatank:
Merged locally with squash.

PR #555, opened 2023-03-08
TILA-2144 application view view

kieferbonk:


ui/components/application/View.tsx
```diff
@@ -0,0 +1,359 @@
...
+                label={t("application:preview.applicationEvent.begin")}
+                value={applicationEvent.begin || ""}
+              />
+              <StyledLabelValue
+                label={t("application:preview.applicationEvent.end")}
+                value={applicationEvent.end || ""}
+              />
+              <StyledLabelValue
+                label={t("application:preview.applicationEvent.minDuration")}
+                value={formatDuration(applicationEvent.minDuration as string)}
```
joonatank, line 269: String(val) or val ?? "" instead of casting

ui/components/application/View.tsx
```diff
@@ -0,0 +1,359 @@
...
+      </CheckboxContainer>
+      <StyledNotification
+        label={t("application:preview.notification.processing")}
+      >
+        {t("application:preview.notification.body")}
+      </StyledNotification>
+      <ButtonContainer>
+        <BlackButton
+          variant="secondary"
+          onClick={() => router.push(`${applicationsUrl}`)}
```
joonatank, line 350: Why router.push? router.back seems more approapriate, or router.replace. Or is there a case where we want user to hit "Cancel" and then "back" on the browser to get them back to the form.

ui/components/application/View.tsx
```diff
@@ -0,0 +1,359 @@
...
+              />
+              <div />
+            </TwoColumnContainer>
+
+            <FormSubHeading>
+              {t("application:Page1.spacesSubHeading")}
+            </FormSubHeading>
+            <UnitList>
+              {sortBy(applicationEvent.eventReservationUnits, "priority").map(
+                (reservationUnit, index) => {
```
joonatank, line 287: Why is this inline? a component seems more appropriate. It already looks like one with variable definitions and a return.

ui/components/application/View.tsx
```diff
@@ -0,0 +1,359 @@
...
+      }
+    }
+    fetchData();
+
+    return () => {
+      mounted = false;
+    };
+  }, [application]);
+
+  const { t } = useTranslation();
```
joonatank, line 189: This can be combined to start of the component useTranslation

ui/components/application/View.tsx
```diff
@@ -0,0 +1,359 @@
...
+      );
+
+      const fetchedReservationUnits = await Promise.all(
+        reservationUnitIds.map((ru) => getReservationUnit(ru.reservationUnitId))
+      );
+
+      if (mounted) {
+        setReservationUnits(
+          mapArrayById(fetchedReservationUnits) as {
+            [key: number]: ReservationUnit;
```
joonatank, line 171: This looks dangerous. Why the cast?

ui/components/application/View.tsx
```diff
@@ -0,0 +1,359 @@
...
+        heading={t("application:preview.basicInfoSubHeading")}
+        theme="thin"
+      >
+        <ApplicantInfoPreview
+          cities={citiesOptions}
+          application={application}
+        />
+      </Accordion>
+      {application.applicationEvents.map((applicationEvent, i) => {
+        const summaryDataPrimary =
```
joonatank, line 208: This looks like a component, why is it inline?

joonatank (changes requested):

joonatank (approved):

PR #556, opened 2023-03-09
TILA-2137: translations for email template generation

vergama:
Language specific texts moved to i18n-jsons and .mjml to .html conversion updated to replace translation keys using them.
Certain common stuff in the .mjml templates abstracted to partials and couple of variable references fixed with currency formatting pipe added for non-zero prices.
Also added translation of text versions of the email templates in the same way so that these can be generated from the same JSON files as the HTML versions.
Cf. also https://helsinkisolutionoffice.atlassian.net/wiki/spaces/KAN/pages/8178106373/Email-pohjat


kieferbonk (approved):
Looks good

PR #560, opened 2023-03-14
TILA-2399 recurring reservation tweaks

kieferbonk:


ui/components/application/Page1.tsx
```diff
@@ -127,7 +127,13 @@ const Page1 = ({
...
 
       fetchedAgeGroupOptions.sort((a, b) => {
-        return (a.minimum || 0) - (b.minimum || 0);
+        const order = ["1-99"];
+        const strA = `${a.minimum || ""}-${a.maximum || ""}`;
+        const strB = `${b.minimum || ""}-${b.maximum || ""}`;
+
+        return order.indexOf(strA) > -1 || order.indexOf(strB) > -1
+          ? order.indexOf(strA) - order.indexOf(strB)
+          : (a.minimum || 0) - (b.minimum || 0);
```
joonatank, line 136: I don't get it. 
After a few tries still don't even know how we are sorting.
Is there a way to convert it to numbers or do early returns for the special cases where minimum / maximum isn't defined?

joonatank (changes requested):

ui/components/application/Page1.tsx
kieferbonk: Extracted the sort function and wrote some tests to cover logic.

joonatank (approved):

PR #566, opened 2023-03-16
TILA-1848 and TILA-2148 Translations missing

aceViilee:


aceViilee:
> `ui/public/locales/en/reservationApplication.json` `reserveeTypePrefix`: "I **create** the application.." (**two periods**) => "I am creating the application…"
I'll make changes. Straight copy paste from the translation excel. That is what the service provider gave us. Go figure.
> 
> `ui/public/locales/en/reservationUnit.json` `notification.onlyRecurring`: "Can only be booked **periodically**" / errors.json has "seasonally" instead of "periodically". `ui/public/locales/sv/reservationUnit.json` `notification.onlyRecurring`: "Kan bokas endast **periodvist**" / errors.json has "säsongsbokning" instead of "periodvist".
I'll make these changes. Same thing as the previous comment
> 
> "**Space**" for "kohde" (beside "tila") in EN is intentional?

Space should be item. I'll check those through.

PR #594, opened 2023-04-03
TILA-2356 remove cache from queries

kieferbonk:


ui/pages/reservation-unit/[...params].tsx
```diff
@@ -101,6 +101,7 @@ export const getServerSideProps: GetServerSideProps = async ({
       QueryTermsOfUseArgs
     >({
       query: TERMS_OF_USE,
+      fetchPolicy: "no-cache",
```
Comment (joonatank) [inline ui/pages/reservation-unit/[...params].tsx:104]: Can't we use something else here or is the no-cache mandatory?  e.g.
```
  fetchPolicy: 'network-only', // Used for first execution
  nextFetchPolicy: 'cache-first', // Used for subsequent executions
```
The data doesn't seem like it would change that often anyway.
The server side might not be that big of a deal though so if it's necessary then sure.

ui/pages/reservation-unit/[...params].tsx
```diff
@@ -245,6 +249,7 @@ const ReservationUnitReservation = ({
     QueryReservationByPkArgs
   >(GET_RESERVATION, {
     variables: { pk: reservationData?.pk },
+    fetchPolicy: "no-cache",
```
Comment (joonatank) [inline ui/pages/reservation-unit/[...params].tsx:252]: Don't like doing this on the client side. Might cause a ton of network requests if the component gets rerendered.
Is it absolutely necessary?

joonatank (changes requested):

joonatank (approved):

PR #599, opened 2023-04-04
TILA-2387: Display recurring reservation instead of regular reservation

joonatank:
The same view as normal requested/reservation but small changes if the reservation is part of a set.
- Different date-time after the header.
- extra Accordion for the other reservations in the set
- update the recurring list styling

Also some fixes to the time format we use.

### DIFF
--- a/admin-ui/src/component/reservations/requested/RecurringReservationsView.tsx
+++ b/admin-ui/src/component/reservations/requested/RecurringReservationsView.tsx
@@ -1,0 +1,91 @@
+import React from "react";
+import { type ReservationType } from "common/types/gql-types";
+import { useTranslation } from "react-i18next";
+import { useQuery } from "@apollo/client";
+import { format } from "date-fns";
+import { RECURRING_RESERVATION_QUERY } from "./queries";
+import { useNotification } from "../../../context/NotificationContext";
+import ReservationList from "../../ReservationsList";
+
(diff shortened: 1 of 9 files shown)


admin-ui/src/component/reservations/requested/RecurringReservationsView.tsx
```diff
@@ -0,0 +1,79 @@
...
+  type ReservationType,
+} from "common/types/gql-types";
+import { useTranslation } from "react-i18next";
+import { useQuery } from "@apollo/client";
+import { format } from "date-fns";
+import { RECURRING_RESERVATION_QUERY } from "./queries";
+import { useNotification } from "../../../context/NotificationContext";
+import ReservationList from "../../ReservationsList";
+
+const StyledHeading = styled.h3`
```
kieferbonk, line 15: For consistency, headings are defined in "common/src/common/typography"

kieferbonk (approved):

admin-ui/src/component/reservations/requested/RecurringReservationsView.tsx
joonatank: Yeah. It's not the same style, it might be semantically table header instead of h3 but the data is not in a table format but a list. The reason to use h3 over div here is just for aria.

PR #608, opened 2023-04-11
TILA-2388: Recurring review fixes

joonatank:
Fix: don't block load more button by default
Fix: end time
Fix: Load more than 100 linked reservations for recurring list

### DIFF
--- a/admin-ui/src/component/reservations/requested/queries.ts
+++ b/admin-ui/src/component/reservations/requested/queries.ts
@@ -1,4 +1,5 @@
 import { gql } from "@apollo/client";
+import { ReservationsReservationStateChoices } from "common/types/gql-types";
 
 export const UPDATE_WORKING_MEMO = gql`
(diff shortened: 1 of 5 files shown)


admin-ui/src/component/reservations/requested/queries.ts
```diff
@@ -108,8 +108,13 @@ export const RESERVATION_QUERY = gql`
 `;
 
 export const RECURRING_RESERVATION_QUERY = gql`
-  query recurringReservation($pk: ID!) {
-    reservations(recurringReservation: $pk, state: ["CONFIRMED", "DENIED"]) {
+  query recurringReservation($pk: ID!, $offset: Int, $count: Int) {
+    reservations(
+      offset: $offset
+      recurringReservation: $pk
+      state: ["CONFIRMED", "DENIED"]
```
kieferbonk, line 115: Api schema enums could be used as the values here.

kieferbonk (approved):

PR #618, opened 2023-04-14
TILA-2390: Deny all button to recurring reservation

joonatank:
Replace the deny / approve buttons with a single Deny All button if the reservation is part of a recurrence.
Deny dialog works with a list of reservations so it can be reused for both use cases.

Apollo caching: cache reservations query because we need 2000 of them in Recurring, mutations update cache no refetching.

### DIFF
--- a/admin-ui/src/component/reservations/requested/hooks/index.ts
+++ b/admin-ui/src/component/reservations/requested/hooks/index.ts
@@ -16,6 +16,7 @@ import {
 import { useNotification } from "../../../../context/NotificationContext";
 import { RESERVATION_DENY_REASONS } from "../queries";
 import { OptionType } from "../../../../common/types";
+import { GQL_MAX_RESULTS_PER_QUERY } from "../../../../common/const";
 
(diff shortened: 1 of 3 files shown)


admin-ui/src/component/reservations/requested/hooks/index.ts
```diff
@@ -67,3 +76,127 @@ export const useReservationData = (
...
+    RECURRING_RESERVATION_QUERY,
+    {
+      skip: !recurringPk,
+      fetchPolicy: "cache-and-network",
+      nextFetchPolicy: "cache-first",
+      errorPolicy: "all",
+      variables: {
+        pk: recurringPk ?? 0,
+        offset: 0,
+        count: limit < LIMIT ? limit : LIMIT,
```
kieferbonk, line 129: Potentially confusing naming - maybe use `defaultOptions.limit` instead of the constant.

kieferbonk (changes requested):

kieferbonk (approved):

PR #623, opened 2023-04-27
TILA-2529 reservation cancellation callback route

kieferbonk:


ui/pages/reservation/cancel.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  useEffect(() => {
+    const { reservationPk } = order || {};
+    if (reservationPk) {
+      deleteReservation({
+        variables: { input: { pk: parseInt(reservationPk, 10) } },
+      });
+    }
+  }, [deleteReservation, order]);
+
```
joonatank, line 60: Can't we run the mutation when user click on a button and redirect to this page after the mutation is success (don't redirect if it fails)?
useEffect for this seem iffy.

ui/components/reservation/DeleteConfirmation.tsx
joonatank, line 1: Confirmation or Cancelled?

joonatank (changes requested):

joonatank (approved):

PR #630, opened 2023-05-02
TILA-2523 hide stepper on paid reservations

kieferbonk:


ui/pages/reservation-unit/[...params].tsx
```diff
@@ -600,19 +600,21 @@ const ReservationUnitReservation = ({
...
-                  onStepClick={(e) => {
-                    const target = e.currentTarget;
-                    const s = target
-                      .getAttribute("data-testid")
-                      .replace("hds-stepper-step-", "");
-                    setStep(parseInt(s, 10));
-                  }}
-                  steps={steps}
-                />
+                {steps.length <= 2 && (
```
Comment (joonatank) [inline ui/pages/reservation-unit/[...params].tsx:603]: This reads like only show the stepper if steps <= 2. 2 steps is a free reservation?
So only show stepper for free reservations?

ui/pages/reservation-unit/[...params].tsx
```diff
@@ -600,19 +600,21 @@ const ReservationUnitReservation = ({
...
+                  <StyledStepper
+                    language={i18n.language}
+                    selectedStep={step}
+                    small={steps.length > 2}
```
Comment (joonatank) [inline ui/pages/reservation-unit/[...params].tsx:607]: small={false} with the above check.

joonatank (changes requested):

ui/pages/reservation-unit/[...params].tsx
Comment (kieferbonk) [reply ui/pages/reservation-unit/[...params].tsx:603]: Oh crap, the title falsely says otherwise. So the objective was to hide it for _paid_ reservations.

PR #637, opened 2023-05-08
TILA-2353: refund reservation

joonatank:


admin-ui/src/component/reservations/requested/DenyDialog.tsx
```diff
@@ -87,6 +178,27 @@ const DialogContent = ({
     },
   });
 
+  const { notifyError, notifySuccess } = useNotification();
+  const { t } = useTranslation();
+
+  const [refundReservationMutation] = useMutation<Mutation>(
```
kieferbonk, line 184: Might be useful to define input/payload types here.

kieferbonk (approved):

admin-ui/src/component/reservations/requested/DenyDialog.tsx
joonatank: Good call

PR #648, opened 2023-05-24
TILA-2056: Add reservation collision check and removing before mutations 

joonatank:


admin-ui/src/component/my-units/MyUnitRecurringReservation/RecurringReservationDone.tsx
```diff
@@ -64,22 +64,35 @@ const btn = [
...
 
-  const failed = props
+  const handleGoToReservation = (id: number) => {
+    const url = `/reservations/${id}`;
+    navigate(url);
+  };
+
+  if (!props.success) {
+    return (
+      <div>ERROR: No data in completed reservation: Should not be here</div>
```
kieferbonk, line 79: Is it intentional to print out errors in english for the end user to see?

kieferbonk (approved):

PR #652, opened 2023-05-29
TILA-2629 ignore buffer on blocked type reservation

kieferbonk:


joonatank (approved):

joonatank (approved):

joonatank (commented):
I'd add `isBlocked` to both RESERVATIONUNIT_RESERVATIONS_FRAGMENT and RESERVATION_COMMON_FRAGMENT  in the admin-ui also, along with the type (admin-ui functions use the type).
Shouldn't be an issue right now, but if there is functions in the common that require isBlocked but do not type check for it, causes weird bugs down the line.

joonatank (approved):

PR #661, opened 2023-06-02
TILA-2638: Permission checks

joonatank:
- Don't allow access to reservation data (links or page) if user has no permissions.
- Allow comments if users own reservation.

### DIFF
--- a/admin-ui/src/component/my-units/ReservationPopupContent.tsx
+++ b/admin-ui/src/component/my-units/ReservationPopupContent.tsx
@@ -7,6 +7,7 @@ import { formatTime } from "../../common/util";
 import { DenseVerticalFlex } from "../../styles/layout";
 import { getReserveeName } from "../reservations/requested/util";
 import { CELL_BORDER } from "./const";
+import VisibleIfPermission from "../reservations/requested/VisibleIfPermission";
 
 const PopupContent = styled.div`
   border: ${CELL_BORDER};
@@ -40,14 +41,19 @@ const ReservationPopupContent = ({
           {formatTime(reservation.begin)} - {formatTime(reservation.end)} /{" "}
           {reservation.reservationUnits?.[0]?.nameFi}
         </Heading>
-        <Reservee>
-          <Link target="_blank" to={reservationUrl(reservation.pk as number)}>
-            {getReserveeName(reservation, 22) || "-"}
-          </Link>
-        </Reservee>
-        {reservation.workingMemo && (
-          <WorkingMemo>{reservation.workingMemo}</WorkingMemo>
-        )}
+        <VisibleIfPermission
+          reservation={reservation}
+          permissionName="can_view_reservations"
+        >
+          <Reservee>
+            <Link target="_blank" to={reservationUrl(reservation.pk as number)}>
+              {getReserveeName(reservation, 22) || "-"}
+            </Link>
+          </Reservee>
+          {reservation.workingMemo && (
+            <WorkingMemo>{reservation.workingMemo}</WorkingMemo>
+          )}
+        </VisibleIfPermission>
       </DenseVerticalFlex>
     </PopupContent>
   );
--- a/admin-ui/src/component/reservations/requested/RequestedReservation.tsx
+++ b/admin-ui/src/component/reservations/requested/RequestedReservation.tsx
@@ -363,37 +363,16 @@ const WorkingMemo = ({
   );
 };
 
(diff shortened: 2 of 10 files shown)


admin-ui/src/component/reservations/requested/VisibleIfPermission.tsx
```diff
@@ -1,49 +1,30 @@
...
-  }
-  if (props.serviceSectorPks.length === 0) {
-    return null;
+  if (!hasPermission(reservation, permissionName)) {
+    // eslint-disable-next-line react/jsx-no-useless-fragment
+    return otherwise ? <>{otherwise}</> : null;
   }
 
-  return <VisibleIfPermission {...props} unitPk={unitPk} />;
+  // eslint-disable-next-line react/jsx-no-useless-fragment
```
kieferbonk, line 26: You might want to set `allowExpressions` to `true` in the eslint rule.

admin-ui/src/component/my-units/ReservationPopupContent.tsx
```diff
@@ -40,14 +41,19 @@ const ReservationPopupContent = ({
...
-          <Link target="_blank" to={reservationUrl(reservation.pk as number)}>
-            {getReserveeName(reservation, 22) || "-"}
-          </Link>
-        </Reservee>
-        {reservation.workingMemo && (
-          <WorkingMemo>{reservation.workingMemo}</WorkingMemo>
-        )}
+        <VisibleIfPermission
+          reservation={reservation}
+          permissionName="can_view_reservations"
```
kieferbonk, line 46: I might consider combining the permission definitions in an enum.

admin-ui/src/component/my-units/ReservationPopupContent.tsx
```diff
@@ -40,14 +41,19 @@ const ReservationPopupContent = ({
...
+        >
+          <Reservee>
+            <Link target="_blank" to={reservationUrl(reservation.pk as number)}>
```
kieferbonk, line 49: Would be best to explicitly convert the type.

admin-ui/src/component/reservations/requested/RequestedReservation.tsx
```diff
@@ -330,49 +299,80 @@ const TimeBlock = ({
...
+  };
 
-  const ref = useRef<HTMLDivElement>(null);
+  return (
+    <>
+      <TextArea
+        label={t("RequestedReservation.workingMemoLabel")}
+        id="workingMemo"
+        helperText={t("RequestedReservation.workingMemoHelperText")}
+        value={workingMemo}
```
kieferbonk, line 347: You might want to sanitize the input at some point.

admin-ui/src/component/reservations/requested/VisibleIfPermission.tsx
```diff
@@ -1,49 +1,30 @@
...
-const VisibleIfPermissionWrapper = (props: {
-  unitPk?: number;
-  serviceSectorPks: number[];
-  permissionName: string;
+}: {
+  reservation: ReservationType;
+  permissionName:
+    | "can_manage_reservations"
+    | "can_comment_reservations"
+    | "can_view_reservations";
```
kieferbonk, line 15: Should we consider a case where the user would need a combination of multiple permissions to access content?

admin-ui/src/component/reservations/requested/RequestedReservation.tsx
joonatank: Not sure if sanitization would make a difference.
Afaik React doesn't allow setting html without dangerouslySetInnerHTML. 

Quick test if it's possible to execute code through the memo textarea:
```
</textarea><script>console.log('baz')</script>
```
On a HTML playground this executes the JS (so saving it in database and executing on another users browser is bad). Here it does not execute which is what I expected since React should not allow unescaped html without dangerouslySetInnerHTML.

admin-ui/src/component/reservations/requested/VisibleIfPermission.tsx
joonatank: I think this component is a temporary solution, if there is direct requirement for multiple permission I'll add it then.
The proper solution would be to refactor the AuthContext and move the new permission check hook there, then maybe improve the VisibleIfPermission component if the hook isn't good enough.
The context uses reducer for data and an outdated Auth package so trying not to touch it too much :S

kieferbonk (approved):

PR #675, opened 2023-06-14
TILA-2675: show the reservation specific buffers in calendars

joonatank:


admin-ui/src/component/my-units/eventStyleGetter.ts
```diff
@@ -82,6 +82,8 @@ const eventStyleGetter =
 
     const isClosed = event?.type === "blocked";
     const isStaff = event?.type === "staff";
+    // @ts-expect-error: TODO: we are dynamically overriding an enum upstream
+    const isBuffer = event?.state === "BUFFER";
```
kieferbonk, line 86: These values should be available in gql typing.

kieferbonk (approved):

PR #684, opened 2023-06-21
TILA-2689: sticky header styling and less buttons 

joonatank:


admin-ui/src/component/StickyHeader.tsx
kieferbonk, line 1: I would consider using only min-* mq's, unless with a strong reason.

kieferbonk (approved):

kieferbonk (approved):

PR #718, opened 2023-08-10
TILA-2703: lists ageGroups in the correct order

vincit-matu:


ui/pages/reservation-unit/[...params].tsx
```diff
@@ -376,23 +376,27 @@ const ReservationUnitReservationWithReservationProp = ({
...
       purpose: reservationPurposes.map((purpose) => ({
         label: getTranslation(purpose, "name"),
         value: purpose.pk,
       })),
-      ageGroup: ageGroups.map((ageGroup) => ({
-        label: `${ageGroup.minimum} - ${ageGroup.maximum}`,
-        value: ageGroup.pk,
-      })),
+      // the sortedAgeGroups array has "1 - 99" as the first element, so let's move it to the end for correct order
+      ageGroup: [...sortedAgeGroups.slice(1), sortedAgeGroups[0]].map(
```
Comment (joonatank) [inline ui/pages/reservation-unit/[...params].tsx:388]: Not sure if it's guarantied anywhere that ageGroups.length > 0.

So I'd prefer [...arr.slice(1), ...arr.slice(0, 1)] or other null checks because
```
const arr = []
[arr.slice(1), arr[0]]  === [undefined]
[...arr.slice(1), ...arr.slice(0, 1)] === []
```

joonatank (changes requested):

ui/pages/reservation-unit/[...params].tsx
Comment (vincit-matu) [reply ui/pages/reservation-unit/[...params].tsx]: I added a null check, which gives a `console.warn()` if there are no ageGroups, and cleaned up the array construction as per your example.

joonatank (approved):
LGTM

PR #720, opened 2023-08-14
Tila 2315: Adds IconButton-component

vincit-matu:
The component can use either a href (and thus be rendered as either a `<Link>` or `<a>`) or a callback argument, so it can be used both as a link and a general CTA button.


joonatank (approved):
LGTM

PR #729, opened 2023-08-21
TILA-2764/cypress env fixes

joonatank:
Run cypress test mock server in port 4000.
Cleanup the tests a bit (reduce timout, remove unused env).
Fix one of the tests that was flaking.


vincit-matu (approved):
LGTM

PR #730, opened 2023-08-21
TILA-2755/show deny reason in summary

joonatank:
Show the cancel / deny reason from the select, not the open comment field.

Refs 2755, 2403


vincit-matu (approved):
LGTM 🚀

PR #737, opened 2023-08-26
TILA-2433: Adds ShowAllContainer, an expanding/accordion-like component

vincit-matu:
ShowAllContainer wraps _n_ child-elements, and renders by default `maximumNumber` of them, followed by a "Show all"-button. The button toggles whether to show all of the child-elements or only `maximumNumber` of them.

It can be used to completely hide child elements by using `0` as `maximumNumber`.

Also creates `common/src/components` for common components, and moves ShowAllContainer and IconButton there for easier use on both admin and user sides of the UI.


joonatank (approved):
LGTM

PR #740, opened 2023-08-28
TILA-1276: turborepo

joonatank:


apps/admin-ui/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+FROM base AS builder
+WORKDIR /app
+RUN npm install -g turbo
+COPY . .
+RUN turbo prune --scope=admin-ui --docker
+
+# Add lockfile and package.json's of isolated subworkspace
+FROM base AS installer
+WORKDIR /app
+# First install the dependencies (as they change less often)
```
matti-lamppu, line 12: If I understand this correctly, this comment indicates that the below `pnpm install --frozen-lockfile` will benefit from docker layer caching, since if only the app changes then we don't need to rebuild dependencies?

This is not the case currently, as on line 6 we do `COPY . .`, which will copy all files in the project to the `builder`, invalidating the layer cache at that point. Then on line 14 and 15 we copy things from the builder, which will invalidates `installer`'s layer cache at that point, triggering a rebuild of dependencies.

You can test this by changing something in `apps/admin-ui/src/App.tsx` for instance. If you want better layer caching, I think you need to separate the building of dependencies to its own stage, or somehow copy less files, or at a later time.

apps/admin-ui/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+COPY turbo.json turbo.json
+ENV SKIP_ENV_VALIDATION=true
+RUN pnpm turbo run build --filter=admin-ui...
+
+FROM base AS runner
+WORKDIR /app
+
+# Don't run production as root
+RUN addgroup --system --gid 1001 nodejs
+RUN adduser --system --uid 1001 nextjs
```
matti-lamppu, line 30: Nitpick: These could be a single line (creates one less layer in the final image):

```dockerfile
RUN addgroup --system --gid 1001 nodejs \
    && adduser --system --uid 1001 nextjs
```

.gitignore
matti-lamppu, line 1: `**/.next` should also be added here.

apps/ui/Dockerfile
matti-lamppu, line 1: Since this Dokcerfile is almost identical to the admin UI, you could consider using an [ARG](https://docs.docker.com/engine/reference/builder/#arg) to set the differing folders at build time, etc. You could benefit from layer caching between the two UIs locally as well.

matti-lamppu:
This is one big pull request! Had a hard time even loading the diffs! 😄 But that's to be expected with the changes.

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,203 @@
...
+          cache: 'pnpm'
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+      - name: Lint admin ui
+        run: |
+          pnpm lint --filter=admin-ui...
+          pnpm lint:css --filter=admin-ui...
+          # TODO run tsc also when it's not broken
+
+  build-admin-ui:
```
matti-lamppu, line 43: For Docker build pipelines, you could use the ready-made Docker actions and push the UI containers to [ghcr.io](ghcr.io). [See how I've done it in the backend](https://github.com/City-of-Helsinki/tilavarauspalvelu-core/blob/main/.github/workflows/build.yml). This allows you to store the layer cache in the registry between builds.

Note that the registry needs to be created manually on the first time for the project by pushing there, then github actions can push automatically.

Dockerfile
```diff
@@ -0,0 +1,47 @@
+FROM node:18-bullseye-slim AS base
+
+# This doesn invalidate layer cache if the result from turbo prune stays the same
+# this step generates a trimmed down version of the package list for installer
+FROM base AS builder
+ARG APP
```
matti-lamppu, line 6: You can declare the `APP` ARG on line 2 and then I think it's valid for the whole file. ARG behaves in such a way that cache misses will only occur when the ARG is used with a value it hasn't been used before.

matti-lamppu: @joonatank Actually, this doesn't seem to be the case, see: https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

Somehow backend is able to build using this method locally, but I'm getting errors in Azure Pipelines for missing args.

matti-lamppu: This might be a [BuildKit](https://docs.docker.com/build/buildkit/) vs OpenShift thing.

vincit-matu (approved):
One heck of a PR :D

apps/admin-ui/Dockerfile
matti-lamppu: This comment seems old, so closing as resolved. Will look through the newer Dockerfile again.

apps/admin-ui/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+WORKDIR /app
+RUN npm install -g turbo
+COPY . .
+RUN turbo prune --scope=admin-ui --docker
+
+# Add lockfile and package.json's of isolated subworkspace
+FROM base AS installer
+WORKDIR /app
+# First install the dependencies (as they change less often)
+COPY .gitignore .gitignore
```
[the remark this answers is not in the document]
matti-lamppu: Resolved

.gitignore
matti-lamppu: This seems to be missing still @joonatank

apps/ui/Dockerfile
matti-lamppu: Dockerfiles combined

.gitignore
joonatank: Was in the sub folders, moved it up.

PR #743, opened 2023-08-29
TILA-2595: fix don't check buffers for BLOCKED

joonatank:
Reuse the single staff reservation logic for recurring reservations, so if either the existing or the new reservation is BLOCKED we don't need to include the buffers in collisions.


vincit-matu (approved):
LGTM 👍

PR #747, opened 2023-08-31
TILA-2535: fix (admin): 502 error because of request size

joonatank:
Use an api endpoint to create the actual fetch request instead of doing it on the client.

Removes unused cookies and headers when making the request.


vincit-matu (approved):
Tested locally: LGTM 🚀

matti-lamppu (approved):

PR #758, opened 2023-09-07
Tila 1528: Adds NotificationList component to common, implements it on ui-side

vincit-matu:


common/src/components/NotificationsList.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const displayedNotificationsList = groupedNotificationsList?.filter(
+    (item) => {
+      if (
+        !closedNotificationsList ||
+        closedNotificationsList.includes(String(item?.id)) ||
+        (item?.target !== target && item?.target !== "ALL")
+      ) {
+        return false;
+      }
+      return true;
```
joonatank, line 147: `return (!closedNotif...)`? or with just: `(item => (!closedNotif...)`

common/src/components/NotificationsList.tsx
```diff
@@ -0,0 +1,167 @@
...
+  let notificationType;
+  switch (notification.level) {
+    case "EXCEPTION":
+      notificationType = "alert" as NotificationType;
+      break;
+    case "WARNING":
+      notificationType = "error" as NotificationType;
+      break;
+    default:
+      notificationType = "info" as NotificationType;
```
joonatank, line 82: Does `let notificationType: NotificationType = 'info'` or `notificationType = 'error' as const` remove the need for `as` cast?

joonatank (approved):

common/src/components/NotificationsList.tsx
vincit-matu: Haha! *facepalm* Indeed.. this became the much more concise:
`(item) =>
      !(closedNotificationsList as string[]).includes(String(item?.id)) ||
      !(item?.target !== target && item?.target !== "ALL")`

common/src/components/NotificationsList.tsx
vincit-matu: Yep, indeed it does. Much cleaner to declare the type than to cast the variables - cheers!

PR #759, opened 2023-09-08
Tila 1585: adds a NotificationsList to the admin side

vincit-matu:


vincit-matu:
wrong ticket number in branch

PR #766, opened 2023-09-15
TILA-1585: use apollo caches

joonatank:
Fix: notifications query broken for regular users
Add: Apollo caches for admin notification
Fix: Invalidate admin notifications caches when mutated
Fix: form validation schema (dates + times validation)
Fix: update mutation didn't report errors
Fix: styling and translations based on UI review


vincit-matu (commented):
Looks good code-wise. 
But it seems I can't edit any notifications. Clicking on any of the existing notifications (on the admin side, where you manage the notifications) brings you to a page stating that said notification does not exist.
Not sure if it's related to the direct aim of this PR, but it seems to break that feature as a side-effect.

vincit-matu (approved):
Now seems to work like it's supposed to. 🚀

PR #767, opened 2023-09-18
TILA-2830: fix: broken multipart form submit

joonatank:


joonatank:
Authentication changes makes this unnecessary (and this conflicts with them). Closing and checking after it if this needs a fix.

PR #768, opened 2023-09-18
TILA-2826: fix: use separate details instead of workingMemo

joonatank:


apps/admin-ui/src/component/reservations/requested/RequestedReservation.tsx
```diff
@@ -233,6 +235,8 @@ const ReservationSummary = ({
           v: reservation?.denyReason?.reasonFi || "-",
         }
       : undefined;
+  // console.log('cancel details: ', reservation.cancelDetails);
+  console.log("handling details: ", reservation.handlingDetails);
```
vincit-matu, line 239: Is this log left intentionally? If so, it should be marked as such for the linter. ;)

vincit-matu (changes requested):
Mostly in order - a bit of nitpicking about a console.log, but nothing major.

vincit-matu (approved):
LGTM

PR #769, opened 2023-09-18
Tila 1585: Display notifications correctly

vincit-matu:
FIxes to the notifications, as requested.

- notification contents displayed as HTML correctly
- close button repositioned
- notification levels fixed, both display/color-wise and the prioritisation of notificatinos
- show correct month (was showing current month - 1)


joonatank (approved):
LGTM

PR #770, opened 2023-09-19
fix: missing username in ui

joonatank:


vincit-matu (approved):
LGTM. Some tests failing, but doesn't seem relevant to the changes made here..

PR #776, opened 2023-09-25
TILA-2723: add: authentication localhost graphql hack

joonatank:
add: a fetch hack to use ip instead of localhost (env ENABLE_FETCH_HACK) for SSR
add: some env validations for API_URL
remove: unused consts


vincit-matu (approved):
all good! 👍

PR #779, opened 2023-09-26
fix: rest api not working with SSR

joonatank:


vincit-matu (approved):
Fixes the SSR problem 👍

PR #784, opened 2023-09-27
TILA-2723: fix: missing csrf token

joonatank:
Fix: missing CSRF token in REST requests caused mutations not to work.
Add: CSRF token to GRAPQHL requests so backend can enable it.


vincit-matu (approved):
LGTM

PR #785, opened 2023-09-28
TILA-1585: fix notifications validators, styling, lang

joonatank:
Fix: Admin styling: remove extra bracket.
Fix: UI styling: padding around text so the close button doesn't overlap.
Fix: Admin validators: validate html text content length without tags.
Fix: UI: fallback to Fi translation if the actual translation is missing.


vincit-matu (approved):
The only thing that comes to mind is that rather trivial language options typing.
But that's nitpicking and possibly not relevant yet. All in all looks good to merge. 👍

PR #788, opened 2023-09-29
TILA-2417: IconButton is now a <button>, if there is an onClick but no href given

vincit-matu:
Now there is no <a> rendered, if the `onClick` attribute returns true, and the href is `javascript:void(0)` (=== the fallback value, aka no `href` was specified).


joonatank (approved):
LGTM

PR #789, opened 2023-10-01
TILA-2812: Fixes the notification z-index and opacity

vincit-matu:
Implemented into the StyledNotification component in styles/utils.ts. Also fixed ReservationUnitEditor.tsx, which was using a custom way for to display notifications, where the vanilla HDS Notification component was in use. It now uses the same StyledNotification which should be used throughout (at least until we unite the notifications of both the ui- and admin-ui sides. There might be some other views which have the same issue.

### DIFF
--- a/apps/admin-ui/src/styles/util.ts
+++ b/apps/admin-ui/src/styles/util.ts
@@ -147,7 +147,7 @@ export const InlineRowLink = styled(BasicLink).attrs({
 
 export const StyledHDSNavigation = styled(Navigation)`
   --breakpoint-xl: 9000px;
-  z-index: var(--tilavaraus-admin-stack-main-menu);
+  z-index: var(--tilavaraus-admin-stack-order-toast);
   .btn-logout {
     span {
       margin: 0;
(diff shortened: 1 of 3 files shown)


apps/admin-ui/src/styles/util.ts
```diff
@@ -147,7 +147,7 @@ export const InlineRowLink = styled(BasicLink).attrs({
 
 export const StyledHDSNavigation = styled(Navigation)`
   --breakpoint-xl: 9000px;
-  z-index: var(--tilavaraus-admin-stack-main-menu);
+  z-index: var(--tilavaraus-admin-stack-order-toast);
```
joonatank, line 150: Is this correct? it's the navigation, not the toast, and admin-stack-order-toast doesn't exist.
Could drop stack-main-menu to 100 (instead of 500) like --tilavaraus-stack-order-navigation in the ui.

joonatank (changes requested):

apps/admin-ui/src/styles/util.ts
vincit-matu: Right you are! Changed, and tested to work.

joonatank (approved):
LGTM

PR #791, opened 2023-10-02
TILA-2817: refactors ReturnLinkList back to a component within the importing files

vincit-matu:
As that list of links isn't reused per se anywhere, both confirmation and cancellation now have their own versions of ReturnLinkList with the correct links in their respective files


apps/ui/components/reservation/ReservationCancellation.tsx
```diff
@@ -182,7 +217,9 @@ const ReservationCancellation = ({ id }: Props): JSX.Element => {
   }, [register]);
 
   const bylineContent = useMemo(() => {
-    const reservationUnit = reservation?.reservationUnits[0];
+    const reservationUnit = reservation?.reservationUnits
+      ? reservation?.reservationUnits[0]
+      : null;
```
joonatank, line 222: `reservation?.reservationUnits?.[0]` works also

apps/ui/components/reservation/ReservationCancellation.tsx
```diff
@@ -198,19 +235,20 @@ const ReservationCancellation = ({ id }: Props): JSX.Element => {
     return <Spinner />;
   }
 
-  const reservationUnit = reservation.reservationUnits[0];
+  const reservationUnit =
+    reservation.reservationUnits && reservation.reservationUnits[0];
```
joonatank, line 239: This might be cause a Typescript error with TS 5.2 because `reservationUnit: boolean | ReservationType` => boolean is not assignable to ReservationType down the line. `Ternary` or `?.` is preferable so it's `ReservationType | undefined`

joonatank (approved):
LGTM

joonatank (approved):

apps/ui/components/reservation/ReservationCancellation.tsx
vincit-matu: That's true! 😅

apps/ui/components/reservation/ReservationCancellation.tsx
vincit-matu: Yeah, I ended up using `reservation?.reservationUnits?.[0]` here too. 👍

PR #796, opened 2023-10-03
Add: typecheck action to pipeline

joonatank:
Only admin and common for now, because types are broken in ui.


vincit-matu (approved):
LGTM

PR #798, opened 2023-10-03
TILA-2840: (admin) replace rest with gql

joonatank:


vincit-matu (approved):
fix commented small stuff and then ok to merge

PR #802, opened 2023-10-06
fix (admin): notification toast auto close

joonatank:
Fix autoClose / dismissable toasts that got broken because of previous changes.
Remove unused autoClose override settings.
Remove custom autoClose using a timer (use the one in HDS instead).
All toasts are dismissable.
Success toasts are auto closed, error toasts are not.


vincit-matu (approved):
We need to do a comprehensive look into the way notifications are handled, and find a solution which would fix the styling issues in one central place - as using inline styling to override the "mystery definitions" isn't an ideal solution either. But after our discussion I think we found agreement in the matter, so this solution is OK for now. 
Good job fixing the notifications forcing the close-icon on top of the text - they look much better, summa summarum! 👍

PR #804, opened 2023-10-09
TILA-2869: fix (ui): redirect on protected routes

joonatank:
Routes that require user to be logged in are redirected to login page on server side.

### DIFF
--- a/apps/ui/public/locales/fi/errors.json
+++ b/apps/ui/public/locales/fi/errors.json
@@ -5,6 +5,7 @@
   "500": {
     "body": "Jotain meni pieleen"
   },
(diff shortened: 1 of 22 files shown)


apps/ui/public/locales/sv/errors.json
joonatank, line 1: Translate!

apps/ui/public/locales/fi/errors.json
```diff
@@ -5,6 +5,7 @@
   "500": {
     "body": "Jotain meni pieleen"
   },
+  "notauthenticated": "Kirjaudu sisään",
```
joonatank, line 8: Change a bit. ...ensin.

PR #805, opened 2023-10-11
fix: (ui) typescript

joonatank:
Use strickNullChecking for Typescript.
Run type checks on CI during lint.
Remove unused component and meaningless tests.
Update commands and README.

Refs: TILA-2424


apps/ui/pages/reservation-unit/[...params].tsx
```diff
@@ -248,7 +235,8 @@ const ReservationUnitReservationWithReservationProp = ({
     "pending"
   );
   const [step, setStep] = useState(0);
-  const [reservation, setReservation] = useState<Reservation | null>(
+  // TODO remove this (it uses an old Rest type)
```
Comment (joonatank) [inline apps/ui/pages/reservation-unit/[...params].tsx:238]: REmove comment

vincit-matu (approved):
Fix the stuff we noticed in our walkthrough, but otherwise LGTM.

PR #809, opened 2023-10-13
fix (ui): Koros styled component broke stylesheet

joonatank:
Incorrect use of variables broke the SSR style sheet generation.


vincit-matu (approved):
lookin good to merge

PR #810, opened 2023-10-16
fix applicant name

joonatank:
Fix: show organisation name if applicable
Refactor: remove useMemo and unnecessary div

Refs: TILA-2829


vincit-matu (approved):
LGTM

PR #814, opened 2023-10-19
TILA-2893: adds type to own reservations gql queries

vincit-matu:
Checks that `reservation.type` is `ReservationsReservationTypeChoices.Normal` in the listings mentioned in TILA-2893.


joonatank (approved):
LGTM 👍

PR #816, opened 2023-10-22
fix: navigation dropdown doesn't go outside the page

vincit-matu:
- aligns the navigation dropdown to the button's right edge instead of left
- doesn't touch mobile styles

Ref: TILA-2819


joonatank (approved):
LGTM

PR #819, opened 2023-10-23
Fix empty-button not showing with the filter tags

vincit-matu:
Use array filter method instead of lodash-omit


joonatank (approved):
LGTM

PR #820, opened 2023-10-24
Fix allocation tabs and query filter

joonatank:
fix: Show reservation unit tabs (buttons without content)
fix: Don't include drafts or cancels in the allocation query

Refs: TILA-2769


vincit-matu (approved):
LeGiT, Mate.

PR #823, opened 2023-10-26
remove (ui): unused open time periods

joonatank:
This is going to be removed from the backend.
Most of the removed code was doing nothing any case, just passing the periods around.


vincit-matu (approved):
I think Eemeli mentioned that he noticed openingHours.ts not actually being used anywhere in the app, too. 👍

PR #827, opened 2023-10-31
fix: don't show units without permissions

joonatank:
Filter out units that the user is not supposed to see after the query, these are returned on purpose for other functionaility but should not be displayed in the ui.

Refs: TILA-2507


vincit-matu (approved):
LGTM 👍

PR #836, opened 2023-11-08
fix: calendar day localization

joonatank:
- Fix calendar always showing days in Finnish.
- Set the calendar localization to user language (not fi).
- Add type safe conversion function from locale -> what we have translated.


vincit-matu (approved):
LGTM 🪨

PR #837, opened 2023-11-08
Fix missing confirmation translations

joonatank:
Fix: incorrect null filtering (doesn't work for undefined)
Fix: missing translations on confirmation page (link etc.)
Fix: rework confirmation page to be separate next page instead of a temporary React state in [...params]


vincit-matu (approved):
LGTM

PR #839, opened 2023-11-09
fix: null checks match undefined also

joonatank:
Fix: null checks not matching undefined
Lint rule: null check needs to be == so it always matches undefined also.


vincit-matu (approved):
LGTM

PR #841, opened 2023-11-09
TILA-2933: fix general terms link

vincit-matu:
Fixes the old links to generalTerms with the proper href for the terms, as they're now known as boooking and not general


joonatank (approved):
LGTM

PR #843, opened 2023-11-10
fix: Increase BlackButton style specificity to get rid of !important

vincit-matu:
Increase BlackButton style specificity to get style definitions to stick, and to get rid of !important

REF: TILA-2779


joonatank (approved):
LGTM

PR #845, opened 2023-11-10
Fix lockfile and gql types

joonatank:
Fix: builds failing because of missing dependency in lock file.
Fix: update to newest GraphQL schema and types
Add: repo root command `pnpm codegen` to update gql-types
Remove: broken cypress e2e tests from github automatic running

Refs: TILA-2930


vincit-matu (approved):
👍

PR #848, opened 2023-11-13
Fix equipments+generic

vincit-matu:
Fixes equipments-label in Swedish and refactors all remaining `generic1` to `booking`.


joonatank (approved):
LGTM

PR #849, opened 2023-11-15
Fix reserving with zero participants

vincit-matu:
Warnings/validation works like it should, but it was a case where minimum number of participants want's specified. Fixed by making minimum participants 1 if it is not specified, or if it has a value of 0 (as there always needs to be at least one participant).
Ref: TILA-2837


joonatank (approved):
LGTM 👍

PR #853, opened 2023-11-16
fix: filter users reservations on backend

joonatank:
Because there is no pagination users were shown zero reservations on the frontend when they had a lot of reservations.

Replace frontend filtering with query filters. Every reservation tab has it's own query with it's own query params. Automatically cache tab queries.

Refs: TILA-2893


vincit-matu (approved):
LGTM

PR #854, opened 2023-11-21
Adds opening hours to recurring reservation time selection

vincit-matu:
Adds handling, styling and aria-fixes to the second page when making recurring reservations. The aria-labels now include the day and time of each slot. The label texts for opening hours are yet to be finalised, but should be added along with the updated legend (now it doesn't contain descriptions for open/not open times, only for the primary and secondary selections).

### DIFF
--- a/apps/ui/components/application/Page2.tsx
+++ b/apps/ui/components/application/Page2.tsx
@@ -4,10 +4,7 @@ import { useTranslation } from "next-i18next";
 import { useRouter } from "next/router";
 import styled from "styled-components";
 import { useFormContext } from "react-hook-form";
-import type {
-  ApplicationEvent,
-  ApplicationEventSchedulePriority,
-} from "common/types/common";
+import type { ApplicationEventSchedulePriority } from "common/types/common";
 import type {
   ApplicationEventScheduleNode,
   ApplicationNode,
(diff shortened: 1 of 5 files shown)


joonatank (commented):
Have to rework the `ApplicationEventSchedulePriority` <-> backend `priority` conversion and the Cell priority types.

They used to be `100 | 200 | 300` not `number` that's why there were consts and casts.
Better would be to add two conversion functions (`transform/convert`?) that change from frontend types to backend types.

apps/ui/components/application/Page2.tsx
```diff
@@ -318,7 +316,14 @@ const Page2 = ({ application, onNext }: Props): JSX.Element => {
     const selectedAppEvents = selectorData
       .map((cell) => cellsToApplicationEventSchedules(cell))
       .map((aes) =>
-        aes.filter((ae) => ae.priority === 300 || ae.priority === 200)
+        aes
+          .filter((ae) => ae.priority === 300 || ae.priority === 200)
+          .map((ae) => {
+            return {
+              ...ae,
+              priority: ae.priority === 300 ? (300 as const) : (200 as const),
```
joonatank, line 324: Remove?

joonatank (approved):

PR #861, opened 2023-11-24
Fix max days broke unit tests

joonatank:
- add Typescript checks to unit tests
- fix: 0 max days before days failing a test 

Ref: TILA-2745


vincit-matu (approved):
LGTM

PR #862, opened 2023-11-24
Fix no reservation unit text

joonatank:
- fix: incorrect text when unit has no reservation units
- refactor: unit page to use same layout components as other pages

Ref: TILA-2867


vincit-matu (approved):
LGTM 🚀

PR #867, opened 2023-11-27
Categorically fix button colors

vincit-matu:
Overwrites the css-variables which hds-buttons use, to have correct theming on our primary and secondary buttons


joonatank (approved):
LGTM 👍

PR #872, opened 2023-11-29
fix: check working memo mutation errors

joonatank:
Wasn't checking GraphQL field errors at all so any request that was valid GQL was reported successful to the user, regardless if it was successful in the backend.


vincit-matu (approved):
LGTM 🚀

PR #874, opened 2023-11-30
fix: invalid date causing browser to freeze

joonatank:
- break the infinite loop
- restrict date selection to only valid days (can be bypassed with keyboard input)


apps/ui/components/reservation-unit/QuickReservation.tsx
```diff
@@ -459,6 +465,23 @@ const QuickReservation = ({
     return null;
   }
 
+  const lastPossibleReservationDate =
+    reservationUnit.maxReservationDuration != null &&
```
vergama, line 469: Should it be `reservationUnit.reservationsMaxDaysBefore` instead of `reservationUnit.maxReservationDuration` ?

PR #876, opened 2023-11-30
fix: unselectable interval and space change effect

joonatank:
fix: Incorrect value for a select.
fix: space change should update the surface and max persons.

Refs: TILA-2816


vincit-matu (approved):
LGTM

PR #877, opened 2023-12-01
fix: dont allow moving reservation units

joonatank:
Using an incorrect unit pk param in the url allowed user to move reservation unit to another unit on save.


vincit-matu (approved):
LGTM

PR #879, opened 2023-12-04
fix: edit reservation unit validation

joonatank:
- allow switching draft <-> published without modification
- don't allow empty html tags without any text
- disallow negative minPerson / maxPerson numbers

Refs: TILA-2816


vincit-matu (approved):
LGTM 👍

PR #884, opened 2023-12-12
Reservation unit card status tags

vincit-matu:
Adds tags about the next possible reservation or lack thereof into search results. Mock data in use, so using `display: hidden` by default so as to not screw up the proper search results, change to `block` to view (originally `inline-flex` but `block` works too).


joonatank (approved):
LGTM 👍

PR #885, opened 2023-12-12
Fix reservation unit time save format

joonatank:
fix: one time field saved in 12h format not 24h
fix: validate min, max duration


vincit-matu (approved):
LGTM 👍

PR #886, opened 2023-12-12
Fix allocation reservation unit selection

joonatank:
- fix: always show all units and reservation-units even if they have no applications
- fix: truncate long reservation-unit names
- fix: tabs overflowing from grid
- refactor: use a common ReservationUnit union type to remove some casts

Refs: TILA-2951


vincit-matu (approved):
LGTM 🚀

PR #887, opened 2023-12-12
Remove the user parameter from payment return url

vincit-matu:
Refs: TILA-2947

### DIFF
--- a/apps/ui/pages/payment-return/index.tsx
+++ b/apps/ui/pages/payment-return/index.tsx
@@ -1,0 +1,56 @@
+import { GetServerSideProps } from "next";
+import { serverSideTranslations } from "next-i18next/serverSideTranslations";
+import { useRouter } from "next/router";
+import { useEffect } from "react";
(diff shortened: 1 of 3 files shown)


joonatank (approved):
LGTM 👍

apps/ui/pages/payment-return/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  );
+  const { data, loading } = useQuery<Query, QueryReservationByPkArgs>(
+    GET_RESERVATION,
+    {
+      variables: { pk: reservationData?.pk },
+      skip: !reservationData?.pk,
+      onError: () => {},
+    }
+  );
+  console.log(data, data?.reservationByPk?.reservationUnits?.[0]);
```
joonatank, line 45: on purpose?

vincit-matu: nope :D

PR #889, opened 2023-12-14
Fixes the price not showing up in quick reservation

vincit-matu:
The duration related fixes in reservationCalendarControls might fix some issues with the calendar, too.


joonatank (approved):
LGTM 👍

PR #891, opened 2023-12-14
fix: breadcrumb use react-router-dom link

joonatank:
fix: No full page refresh anymore when navigating.
refactor: remove search based alias (use array of structs instead).


vincit-matu (approved):
LGTM

PR #894, opened 2023-12-14
Changes durationOptions value to "H:mm" format

vincit-matu:
Makes duration use a H:mm-format string as the value again, instead of a number. Seems to fix booking a time via the calendar, and the quick reservation (w/ price) works, too. However, doesn't yet fix the problems with the calendar showing incorrect reservable times. :/


joonatank (approved):
👍

PR #896, opened 2023-12-18
Fixes reservation unit page calendar & quick reservation

vincit-matu:
Should fix all the problems encountered after the API change.


joonatank (approved):
👍

PR #899, opened 2023-12-19
chore: update gql types from core

joonatank:
Bump gql types to most recent core. 

- fix: ID -> Int change (string -> number)


vincit-matu (approved):
LGTM

PR #901, opened 2023-12-20
Fix cant move reservation

joonatank:
- fix: reservation edit page not matching the reservation creation
- fix: open for 24h not working because it returns only one timespan
- fix: sort users own reservations by next upcoming
- fix: broken ui unit tests
- fix: missing ID -> Int gql type change
- refactor: EditStep0 types and sub components

Refs: TILA-3000


vincit-matu (approved):
LGTM 🎄

PR #902, opened 2023-12-20
fix: profile url env validation and documentation

joonatank:
- Profile url is not required but it has to be an url.
- update README
- add empty default value to example env


vincit-matu (approved):
LGTM

PR #903, opened 2023-12-20
Adds initial allocated events tab to recurring reservations review

vincit-matu:
Adds the tab, but the table doesn't yet list the correct info (as it's a copy of ApplicationEventsDataLoader/Table with changes to the query).


joonatank (approved):
👍

PR #905, opened 2023-12-20
fix: is reservable

joonatank:
remove: incorrect check (reservationBegins can be null)
add: logging if unit is not reservable
fix: combine two query results (just in case since we are making both)


vincit-matu (approved):
🎅

PR #906, opened 2023-12-20
Fix: show-only-available checkbox now works on initial page load

vincit-matu:
The "show only available times" checkbox was checked by default, but the filter wasn't in effect on the initial page load.


joonatank (approved):
👍

PR #923, opened 2024-01-11
fix: separate dateOfBirth query

joonatank:
Fix the component that makes separate requests when user wants to see the birth date.
Only request birth date if explicitly needed by the user.
Remove the extra component, use user pk to make the requests.


apps/admin-ui/src/component/BirthDate.tsx
```diff
@@ -1,41 +1,79 @@
...
+    variables: {
+      pk: userPk,
+    },
+    fetchPolicy: "network-only",
+    skip: !userPk || !visible,
+  });
+
+  const { t } = useTranslation();
+
+  const user = data?.user ?? undefined;
```
ranta, line 53: `?? undefined` seems redundant

ranta (approved):

apps/admin-ui/src/component/BirthDate.tsx
joonatank: It's because GQL types from codegen are Maybe<T> and Maybe !== undefined. So TS is going to cry about it, while in JS they work the same.

joonatank: But in this case it's not needed. Habit of putting it there because it's needed when passing to other components with type signature `var?: T`.

PR #924, opened 2024-01-11
Filter application round

joonatank:
Add new application round filters and sorting using URL as the only state.

- Add: Allocated application events tab.
- Filters and sorts implemented for all three tabs.
- Preserve filter and sort state while switching between tabs.
- Add: apollo caching for queries (making tab switches instant).
- Add: separate isLoading state for Tables (darker colour).
- Fix: don't remove table from DOM when loading new query results (causes page jumping).
- Fix: don't add history when changing filters / sort (allocation page also).
- Fix: application / event status colours to match design.

Refs: TILA-2973, TILA-2986, TILA-2987, TILA-2988, TILA-2982, TILA-2983, TILA-2984


vincit-matu (approved):
LGTM

PR #926, opened 2024-01-15
Add new azure pipeline triggers

matti-lamppu:
## Add pipeline triggers for new Azure pipelines

Should trigger corresponding dev build pipeline when code is merged to the main branch from the appropriate folders. Deploy to the dev environment is triggered automatically after the build finishes.

Should trigger corresponding release build when a tag matching the given rules is created (more info below). Deploy to the test environment is triggered after the build finishes. Staging and production deployments use the release builds, but need to be triggered manually from Azure pipelines (can go over this in more detail later).

Tag rules should work like this:

`v1.0.0` or `release-1.0.0` or `hotfix-1.0.0` -> release UI and Admin UI
`v1.0.0-ui` or `release-1.0.0-ui` or `hotfix-1.0.0-ui` -> release only UI
`v1.0.0-admin` or `release-1.0.0-admin` or `hotfix-1.0.0-admin` -> release only Admin UI

Note: The exclude rule is undocumented, and might not work. Needs testing.

TODO:
- These new triggers need to be linked to the corresponding Azure pipelines after merging to make them work.
- There are still some issues in the pipelines, which I will try to resolve with the backend build, so don't merge before I give the OK.


joonatank (approved):
👍

PR #927, opened 2024-01-15
fix: gql decimal type is string not number

joonatank:
Mutations accept number also so they at least used to work. A lot of the code did implicit string conversions, but some comparisons might have been broken.

Refs: TILA-2999


vincit-matu (approved):
LGTM

PR #928, opened 2024-01-17
rename list components that are not list

joonatank:
Quick reorganisation. Removed the confusing lists component directory containing Table and misc components.


vincit-matu (approved):
👍

PR #931, opened 2024-01-18
fix: double test ids on reservation page

joonatank:
Replace only mobile / desktop dom elements with a responsive grid.

Fixes testid="reservation__reservation-info-card__duration"


vincit-matu (commented):
👍 lgtm

PR #935, opened 2024-01-22
fix: disabling time should disable it's validation

joonatank:
mutations don't save the values anyway, but the form validation wasn't disabled if the field was disabled by the user.


vincit-matu (approved):
LGTM

PR #940, opened 2024-01-24
fix: make docker images reusable

joonatank:
NEXT_PUBLIC_* variables are substituted during build step which makes it impossible to reuse the docker image.
Load them server side and pass them as props to client instead.

Refs: TILA-2809, TILA-3086


vincit-matu (approved):
lgtm

PR #946, opened 2024-01-30
Adds reservation unit selection to recurring application page2

vincit-matu:
Adds reservation unit selection to a season reservation application, so the user can check opening times of different reservation units. 
Made with a simple useState, as it's meant only for showing the time-slots meant for seasonal bookings and doesn't affect the form contents per se. Some prop drilling necessary, to display the reservation unit selector in the right place.

Also fixes some styling issues on page 2 (legend, reset button..).

Refs: TILA-3052

### DIFF
--- a/apps/ui/components/application/Page2.tsx
+++ b/apps/ui/components/application/Page2.tsx
@@ -75,7 +75,7 @@ const applicationEventSchedulesToCells = (
           end: t && +t.end.split(":")[0] === 0 ? 24 : t && +t.end.split(":")[0],
         };
       }) ?? [];
-    // state is 50 if the cell is outside of the opening hours, 100 if it's inside
+    // state is 50 if the cell is outside the opening hours, 100 if it's inside
     for (let i = firstSlotStart; i <= lastSlotStart; i += 1) {
       const isAvailable = dayOpeningHours.some(
         (t) => t.begin != null && t.end != null && t?.begin <= i && t?.end > i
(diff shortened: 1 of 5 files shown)


apps/ui/components/application/Page2.tsx
```diff
@@ -214,15 +214,23 @@ const getApplicationEventsWhichMinDurationsIsNotFulfilled = (
...
   const openingHours = filterNonNullable(
-    application?.applicationEvents?.[0]?.eventReservationUnits?.[0]
-      ?.reservationUnit.applicationRoundTimeSlots
+    application?.applicationEvents?.[0]?.eventReservationUnits?.[
+      selectedReservationUnit
+    ]?.reservationUnit.applicationRoundTimeSlots
+  );
+  const reservationUnitOptions = filterNonNullable(
+    application?.applicationEvents?.[0]?.eventReservationUnits?.map(
+      (n, idx) => ({ value: idx, label: n?.reservationUnit.nameFi ?? "" })
```
joonatank, line 230: I'd use `pk` as value for the selected reservationUnit.
Requires the use of find to map the value, but it's not brittle to cache reset / refetch, and React keys are correct (in case it would dynamically be removed from the array).

apps/ui/components/application/TimeSelector.tsx
```diff
@@ -370,7 +408,16 @@ const TimeSelector = ({
   return (
     <>
       <OptionWrapper>
-        <PrioritySelect
+        <StyledSelect
+          id={`time-selector__select--reservation-unit-${index}`}
+          aria-labelledby={t("appplication:Page2.")}
+          options={reservationUnitOptions}
+          defaultValue={reservationUnitOptions[selectedReservationUnit]}
```
joonatank, line 415: Is there a reason to set `defaultValue` and not `value`?
Technically it shouldn't matter right now, but if it's `defaultValue` I'd rename the `selectedReservationUnit` so it's clear that it can't be controlled after initialisation.

Another issue with `defaultValue` is that Queries are async, so unless the component blocks till we have query data the `defaultValue` will always be undefined.

joonatank (changes requested):
Change or clarify the 2 points

apps/ui/components/application/TimeSelector.tsx
vincit-matu: No real reason. It takes after the priority selector which was already there - `value` is probably better.

apps/ui/components/application/Page2.tsx
vincit-matu: Yep, a more robust way to do it. Refactored.

PR #947, opened 2024-01-30
30min seasonal reservation periods

vincit-matu:
Removes 15min and 45min options from seasonal reservation application

Ref: TILA-3020


joonatank (approved):
👍

PR #948, opened 2024-01-30
fix: page 3 translations

vincit-matu:
Page 3 was displaying page 4 heading and subheading and non-finnish translations didn't show due to erroneous translation key. Fixed.


joonatank (approved):
👍

PR #952, opened 2024-02-01
Add reservation block whole day select

vincit-matu:
Adds a new select input on the reservation unit edit/create page: one can choose to 
1. either have no breaks,
2. have the reservation block the whole day, or
3. specifying breaks.
If 3 is selected the checkboxes and inputs for breaks before and after the reservation are shown below.


apps/admin-ui/src/spa/ReservationUnit/edit/form.ts
```diff
@@ -983,8 +997,9 @@ export function transformReservationUnit(
       hasScheduledPublish && hasPublishEnds
         ? constructApiDate(publishEndsDate, publishEndsTime)
         : null,
-    bufferTimeAfter: hasBufferTimeAfter ? bufferTimeAfter : null,
-    bufferTimeBefore: hasBufferTimeBefore ? bufferTimeBefore : null,
+    reservationBlockWholeDay: reservationBlockWholeDay === "blocks-whole-day",
+    bufferTimeAfter: hasBufferTimeAfter ? bufferTimeAfter : 0,
+    bufferTimeBefore: hasBufferTimeBefore ? bufferTimeBefore : 0,
```
joonatank, line 1002: Could remove the ternary here (it's always a number, 0 is the default value).

apps/admin-ui/src/spa/ReservationUnit/edit/index.tsx
```diff
@@ -153,6 +153,19 @@ const SubAccordion = styled(Accordion)`
   }
 `;
 
+const BufferWrapper = styled.div`
+  display: grid;
+  grid-template-columns: repeat(12, 1fr);
+  grid-column: 2 / 13;
```
joonatank, line 159: isn't `2 / 13` === `2 / -1`? meaning it skips the first column, but fills everything else?
Just a preference, but -1 is better for the last column (and it's currently used in the code).

apps/admin-ui/src/spa/ReservationUnit/edit/index.tsx
```diff
@@ -1071,60 +1084,104 @@ function ReservationUnitSettings({
...
+                      onChange={(e) => onChange(e.target.value)}
+                      checked={value != null && value === "buffer-times-set"}
+                    />
+                  </SelectionGroup>
+                )}
+              />
+            </Span12>
+            {watch("reservationBlockWholeDay") === "buffer-times-set" && (
+              <BufferWrapper>
+                <Span6>
```
joonatank, line 1126: I'd check if you can remove the `Span6` here (they are bad anyway, and create extra divs), if we are defining a custom grid wrapper for it already.

joonatank (approved):
Comments are style things / preference (or further cleanup). Ok without too.

apps/admin-ui/src/spa/ReservationUnit/edit/index.tsx
vincit-matu: Good catch! Yeah, skips first column, and fills out the rest of the space.

apps/admin-ui/src/spa/ReservationUnit/edit/index.tsx
vincit-matu: Yeah, it was enough to define the custom grid wrapper as `grid-template-columns: repeat(2, 1fr);` - fixed now.

PR #953, opened 2024-02-01
refactor: split a hook that does two things

joonatank:
Reduces unneccessary queries if we only need the delete mutation and makes the logic easier to follow.


vincit-matu (approved):
👍

PR #954, opened 2024-02-01
fix: missing fields in reservation query

joonatank:
Fragment queries where easily possible to avoid code duplication, and similar errors in the future.


vincit-matu (approved):
good stuff! 👍

PR #956, opened 2024-02-02
breaking: backend payment order api change

joonatank:
Newest core added the payment order link to reservation instead of separate fields.


vincit-matu (approved):
lgtm

joonatank:
Merge when backend makes a new release build.

Making a release from this on the frontend will break environments.

PR #957, opened 2024-02-02
Fix buffer time query and mutations

joonatank:
- api change: replace null buffer time with 0
- fix: remove buffer change from edit reservation mutation
- fix: show buffers in reservation calendar

Refs: TILA-3078, TILA-3079, TILA-3083, TILA-3084


vincit-matu (approved):
👍

PR #958, opened 2024-02-02
Adds short day of the week to the date, above my-units calendar

vincit-matu:
## 🛠️ Changelog
- Adds the short day name next to the date, above the calendar in my-units section (admin side)
- Also attempts to add the same favicon which is in use on the ui/client-side, in exactly the same way - however it doesn't seem to have intended consequences (..or any at all 😮)

## 🧪 Test plan
- Check out the calendar in a reservation unit page, under my units.
- See if anything sticks out to you regarding why the favicon isn't showing as on the UI-side

## 🎫 Tickets

- TILA-3010


joonatank (changes requested):
Doesn't work, like we talked.  Just marking so it doesn't get merged accidentially.

vincit-matu:
> Doesn't work, like we talked. Just marking so it doesn't get merged accidentially.

Yup, the date input is broken all-in-all. Moved the weekday outside the `input`-element, but it would need design decisions and then styling according to those before being ready.

apps/ui/components/reservation-unit/Head.tsx
joonatank, line 1: Got an extra commit here from the other PR :)

vincit-matu: Hehe, fixed! 😊

PR #961, opened 2024-02-05
fix: empty links on my reservation page

joonatank:
Two links on reservation page were empty because the field was missing.


vincit-matu (approved):
👍

PR #963, opened 2024-02-06
fix: translation typo

joonatank:


apps/admin-ui/src/i18n/messages.ts
```diff
@@ -449,7 +449,7 @@ const translations: ITranslations = {
     applicantTypes: {
       INDIVIDUAL: ["Yksityishenkilö"],
       COMPANY: ["Yritys"],
-      COMMUNITY: ["Yhdistys, rekisteröimäton"],
+      COMMUNITY: ["Yhdistys, rekisteröimätön"],
```
vincit-matu, line 452: What's the change here?

vincit-matu (approved):
I don't see any actual change, but if it's as it should be then LGTM

apps/admin-ui/src/i18n/messages.ts
joonatank: O -> Ö

PR #968, opened 2024-02-07
fix: incorrect applied count

joonatank:
## 🛠️ Changelog
Incorrectly calculated the number of applied events instead of showing how many times the user requested per week.


vincit-matu (approved):
LGTM

PR #969, opened 2024-02-07
comments out buffer options from reservation unit editing for now

vincit-matu:
## 🛠️ Changelog

- Hides the buffer options from the reservation unit create/edit form

## 🧪 Test plan

- Create/edit a reservation unit and notice that the section isn't shown

## 🎫 Tickets

- TILA-3029


joonatank (approved):
👍

PR #970, opened 2024-02-07
fix order of non reservable notification error text checking

vincit-matu:
## 🛠️ Changelog

- Corrects the logic behind what is shown in the notification for a reservation unit not being reservable
- Splits said notification to its own component `NonReservableNotification`, for clearer code

## 🧪 Test plan

- Check that the code seems reasonable.

## 🎫 Tickets

- TILA-3075


apps/ui/components/reservation-unit/Head.tsx
```diff
@@ -107,6 +107,38 @@ const UnitName = styled(H3).attrs({ as: "h2" })`
...
+              date: reservationUnit.reservationBegins
+                ? formatDate(reservationUnit.reservationBegins, "d.M.yyyy")
+                : "",
+              time: reservationUnit.reservationBegins
+                ? formatDate(reservationUnit.reservationBegins, "H.mm")
+                : "",
+            })
+          : "notReservable"
+    }`
+  );
```
joonatank, line 132: This is too hard to follow (nested ternaries). I'd separate the error message generation to a function, but anything other than multiple ternaries (define the messages as variables first, then pick the right one in one select etc.).

joonatank (commented):
It's an improvement, but I can't read the error message generation logic.

apps/ui/components/reservation-unit/Head.tsx
```diff
@@ -107,6 +107,40 @@ const UnitName = styled(H3).attrs({ as: "h2" })`
...
+            ? formatDate(reservationUnit.reservationBegins, "H.mm")
+            : "",
+        })
+      : false;
+  const returnText = t(
+    `reservationUnit:notifications.${
+      reservationUnit.reservationKind ===
+      ReservationUnitsReservationUnitReservationKindChoices.Season
+        ? "onlyRecurring"
+        : getFutureText() || "notReservable"
```
joonatank, line 132: Is the future text supposed to be inside the translation?

vincit-matu: Nope. Refactored the whole thing to a more readable form.

joonatank (approved):
👍

PR #971, opened 2024-02-08
Fix allocation events per week

joonatank:
## 🛠️ Changelog
- fix: incorrect allocated state from backend, skip the check for now 
- fix: incorrectly comparing to unallocated and not to requested eventsPerWeek

## 🎫 Tickets
- TILA-3060


vincit-matu (approved):
👍

PR #972, opened 2024-02-08
Fix protected route middleware

joonatank:
## 🛠️ Changelog
- Fix: login redirect when making a new seasonal reservation application.
- Fix: use SSR to query the possible reservation-units for seasonal reservation.
- Fix: use a Link instead of router.push for next button.

## 🧪 Test plan
- Create a new application for seasonal reservation as a non logged in user.

## 🎫 Tickets

- TILA-####


vincit-matu (approved):
Looks good 🥷

PR #974, opened 2024-02-09
Fix admin single reservation validators

joonatank:
## 🛠️ Changelog
- Fix: empty buffers rendering 0 instead of being hidden in the form.
- Fix: validate the form that reservation can't start before now.

## 🧪 Test plan
- In admin ui try making a reservation using the single reservation modal to a past date or time. 

## 🎫 Tickets
- TILA-3017


vincit-matu (approved):
LGTM

PR #980, opened 2024-02-12
Fix unit permissions getting elevated

joonatank:
## 🛠️ Changelog

- Fix: reduce the number of waiting to be handled shown on the dot (partial).
- Fix: unit permission showing elevated permissions if user had more permissions in another unit.

## 🧪 Test plan

- Waiting to be handled dot should not show any past reservations anymore.
- Waiting to be handled should not show results for units where the user has no access at all.
- Giving a user unit "view" and "all" permissions to separate units should now hide approval / deny buttons correctly (where user only has "view" access).
- Giving a user unit "reserve" and "all" permissions to separate units should block the user from seeing a reservation (where user only has "reserve" permissions).

## 🎫 Tickets

- TILA-3131

### DIFF
--- a/apps/admin-ui/src/common/queries.ts
+++ b/apps/admin-ui/src/common/queries.ts
@@ -173,9 +173,14 @@ export const DELETE_RESOURCE = gql`
   }
 `;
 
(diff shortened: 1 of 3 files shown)


vincit-matu (approved):
LGTM

PR #988, opened 2024-02-14
refactor: add reusable query filter components

joonatank:
## 🛠️ Changelog

- Use directly the url params based on name key.
- Use common translations and paths based on the name.

Requires react-router-dom so can't yet reuse in client application.

## 🧪 Test plan

- For now only used on allocation and application-round/:id pages so check that the filters work and translations are correct.

## 🎫 Tickets

- TILA-3169


vincit-matu (approved):
Good stuff 😎

PR #990, opened 2024-02-14
fix: show blocked reservations as closed time slots

vincit-matu:
## 🛠️ Changelog

- Blocked reservations should show as the reservation unit being closed, so adds a check for `isBlocked` and applies same styling as for when the reservation unit is closed.
- Assumes that we also want to hide the time (and/or other possible text) so sets `color` to `transparent`.

## 🧪 Test plan

- Create an event/a reservation on the admin side which should be seen as blocked on the client side
- Check the reservation unit's page: the reservation should show as closed
- Same thing for EditStep0 (== edit reservation, first step)

## 🎫 Tickets

- TILA-3016

### DIFF
--- a/apps/ui/components/reservation/EditStep0.tsx
+++ b/apps/ui/components/reservation/EditStep0.tsx
@@ -22,13 +22,7 @@ import {
 import classNames from "classnames";
 import { IconArrowRight, IconCross } from "hds-react";
 import { useRouter } from "next/router";
-import React, {
-  CSSProperties,
-  Children,
-  useCallback,
-  useMemo,
-  useState,
-} from "react";
+import React, { Children, useCallback, useMemo, useState } from "react";
 import { useTranslation } from "next-i18next";
 import { useMedia } from "react-use";
 import styled from "styled-components";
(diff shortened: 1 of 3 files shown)


apps/ui/components/reservation/EditStep0.tsx
```diff
@@ -44,6 +38,7 @@ import { BlackButton, MediumButton } from "@/styles/util";
 import Legend from "../calendar/Legend";
 import ReservationCalendarControls from "../calendar/ReservationCalendarControls";
 import { CalendarWrapper } from "../reservation-unit/ReservationUnitStyles";
+import { eventStyleGetter } from "@/pages/reservation-unit/[id]";
```
joonatank, line 41: I'd rather not import from @pages. Could we move it to new or something `components/reservation` or `components/reservationCommon`. Something temporary is fine.

joonatank (commented):
Great, getting rid of the extra `eventStyleGetter` function, but would change the export location.

apps/ui/components/reservation/EditStep0.tsx
vincit-matu: Made a new `calendarUtils.ts` file under `components/common`.

joonatank (approved):
👍

PR #991, opened 2024-02-14
fix: unify the labeling of reservation control fields

vincit-matu:
## 🛠️ Changelog

- Quick reservation used its own set of duplicated translation keys for the field label texts. Changes the referred keys to be the same ones used by the calendar.
- The reset button in the calendar reservation view was styled wrong. Changes it to `secondary` instead of `supplementary` got the basic styling right. Manually fixed the rest.
- Restyles ReservationCalendarControls to make everything fit
- Removes start time input from QuickReservation element, as it was confusing and we list all the available times for the given date anyway right below the input. The user can navigate via the quick reservation time chunks.
- Updates translations for seasonal application unit selection label on Page2

## 🧪 Test plan

- Go to any reservation unit page and select a time, then open the reservation controls accordion:
  - The labels should match between the quick reservation and the calendar reservation controls
  - The calendar reservation controls reset button looks and behaves as it should (== grayscale)
  - The duration options are in "2t 30min"-format, starting from 2h. Any values below that should be displayed in minutes only ("90 min").

## 🎫 Tickets

- TILA-3018

### DIFF
--- a/apps/ui/components/reservation-unit/QuickReservation.tsx
+++ b/apps/ui/components/reservation-unit/QuickReservation.tsx
@@ -581,7 +581,7 @@ const QuickReservation = ({
       <Selects>
         <DateInput
           id="quick-reservation-date"
-          label={t("reservationCalendar:quickReservation.date")}
+          label={t("reservationCalendar:startDate")}
           initialMonth={new Date()}
           language={getLocalizationLang(i18n.language)}
           onChange={(_val, valueAsDate) => {
@@ -609,7 +609,7 @@ const QuickReservation = ({
         <StyledTimeInput
           key={`timeInput-${time}`}
           id="quick-reservation-time"
-          label={t("reservationCalendar:quickReservation.time")}
+          label={t("reservationCalendar:startTime")}
           hoursLabel={t("common:hours")}
           minutesLabel={t("common:minutes")}
           defaultValue={time}
@@ -628,7 +628,7 @@ const QuickReservation = ({
         <StyledSelect
           key={`durationSelect-${duration?.value}`}
           id="quick-reservation-duration"
-          label={t("reservationCalendar:quickReservation.duration")}
+          label={t("reservationCalendar:duration")}
           options={durationOptions}
           onChange={(val: OptionType) => setDuration(val)}
           defaultValue={duration}
--- a/apps/ui/components/reservation-unit/QuickReservation.tsx
+++ b/apps/ui/components/reservation-unit/QuickReservation.tsx
@@ -9,7 +9,7 @@ import {
   isBefore,
   isValid,
 } from "date-fns";
-import { DateInput, IconAngleDown, Select, TimeInput } from "hds-react";
+import { DateInput, IconAngleDown, Select } from "hds-react";
 import { padStart } from "lodash";
 import { useTranslation } from "next-i18next";
 import { useLocalStorage } from "react-use";
@@ -74,6 +74,9 @@ const Price = styled.div`
   }
   padding-bottom: var(--spacing-m);
   height: var(--spacing-m);
+  &:empty {
+    display: none;
+  }
 `;
 
 const Selects = styled.div`
@@ -97,16 +100,6 @@ const Selects = styled.div`
     }
 
     grid-template-columns: 1fr 1fr;
-  }
-`;
-
-const StyledTimeInput = styled(TimeInput)`
-  > div > div {
-    > div {
-      margin: 0;
-      width: var(--spacing-2-xs);
-    }
-    padding: 0 var(--spacing-xs) 0 var(--spacing-xs) !important;
   }
 `;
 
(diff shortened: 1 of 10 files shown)


apps/ui/components/reservation-unit/QuickReservation.tsx
```diff
@@ -286,17 +278,14 @@ type AvailableTimesProps = {
 function getAvailableTimesForDay({
   day,
   duration,
-  time,
   isSlotReservable,
   reservationUnit,
   fromStartOfDay = false,
```
joonatank: `fromStartOfDay` seems useless now

apps/ui/components/reservation-unit/QuickReservation.tsx
```diff
@@ -286,17 +278,14 @@ type AvailableTimesProps = {
...
 }: AvailableTimesProps): string[] {
   if (reservationUnit == null) return [];
   const [durationHours, durationMinutes] =
     duration?.split(":").map(Number) || [];
```
joonatank, line 286: This is bit iffy now that `label !== value` pretty sure it's gonna cause issues later.

apps/ui/components/reservation-unit/QuickReservation.tsx
```diff
@@ -431,7 +421,7 @@ const QuickReservation = ({
     formatDate(nextHour.toISOString(), "HH:mm")
   );
   const [duration, setDuration] = useState<OptionType | undefined>(
-    durationOptions.find((n) => n.value === "1:00") || durationOptions[0]
+    durationOptions.find((n) => n.value === "60 min") ?? durationOptions[0]
   );
```
joonatank, line 424: Value is never `60 mins`. Using a number for the value is the correct solution here.

apps/ui/modules/reservation.ts
```diff
@@ -44,10 +45,21 @@ export const getDurationOptions = (
...
+      if (hms.m !== 0)
+        return t("common:abbreviations.minute", { count: hms.m });
+      return "";
+    };
+    const optionString = `${hourString} ${minuteString()}`;
     timeOptions.push({
-      label: `${hms.h}:${minute}`,
-      value: `${hms.h}:${minute}`,
+      label: optionString,
+      value: `${hms.h}:${String(hms.m).padEnd(2, "0")}`,
```
joonatank, line 62: I'd change the `value` to a number (duration in minutes) now. Introduces too many bugs because of the string comparison and `label !== value`

joonatank (changes requested):

apps/ui/components/reservation-unit/QuickReservation.tsx
vincit-matu: So it was incorrect to start with (as it was using a string in different format)?

vincit-matu: It should use the "1:00" which was there to begin with, methinks. (as the option's `value` is still `1:00` - only labels changed)

apps/ui/modules/reservation.ts
vincit-matu: That's way outside the scope of this ticket in my opinion. Last time I tried that was when I completely fucked up the calendar. I was trying to convert the duration to minutes when the time format changed.

TBH I don't see what bugs would be introduced simply from `label !== value` here.

apps/ui/components/reservation-unit/QuickReservation.tsx
vincit-matu: ...altho I wonder if that entire comparison is actually unnecessary. What we want is `durationOptions[0]` anyway - don't know why try to force the 1h value there if the reservation unit has a different minimum reservation duration.

apps/ui/components/reservation-unit/QuickReservation.tsx
vincit-matu: Hmm... Not sure I agree here. What was asked for was that the durations would be _displayed_ in the same manner everywhere. Changing only the `label` leaves everything else in place logic-wise === everything works. Or "works", but as well as it used to before the change.

As I see it the only change is that the user sees the time in the requested format. Changing the value to minutes (as suggested below) is a good idea, but not necessary for these changes. Or that's my two cents at least.

apps/ui/components/reservation-unit/QuickReservation.tsx
vincit-matu: Yeah, that's why it was removed 😜

apps/ui/modules/reservation.ts
joonatank: Not by that, but both are strings and string comparison causes problems. Accidentially picking `label` instead of `value` and TS isn't gonna catch the problem.
You already did it in another place, accidentially changing value comparison even though values didn't change.

apps/ui/components/reservation-unit/QuickReservation.tsx
vincit-matu: Asked Krista, and we should go with the 1h assumed default time (and not the first option, unless there isn't a valid 1h option).

joonatank (approved):
👍

PR #993, opened 2024-02-15
fix: "begin/end" => "beginDate/endDate" in reservations

vincit-matu:
## 🛠️ Changelog

- The API switched to using `beginDate` and `endDate` (instead of `begin`& `end`) in `reservations` listing endpoint, which breaks the current frontend. This should correct the use of said parameters.
- ❗This breaks the app if PR #1114 hasn't been merged to the backend❗ 

## 🧪 Test plan

- The TS-check should tell about any errors. But in practice once the backend PR is merged, check that pages which use the `reservations` endpoint behave correctly (admin-ui: reservations-page, client-ui: reservations index and edit pages, reservation unit page, common: the calendar)

## 🎫 Tickets

- TILA-3096


joonatank (approved):
👍

PR #1000, opened 2024-02-21
add: loading states to relevant reservation buttons

vincit-matu:
## 🛠️ Changelog

- Adds `isLoading` status to "Varaa" and "Jatka" buttons in the reservation process, and in reservation edit. Also in the "Peru varaus" button when canceling.

## 🧪 Test plan

- Go to any reservation unit page and make a reservation. You should see the "Varaa"-button change to a loading spinner once clicked. The same goes for the "Jatka"-buttons in the different steps when making a reservation.
- Once the reservation is made, go to "Omat varaukset" and edit the reservation. The buttons in the edit process should also have loading states.
- Finally: cancel the reservation you just made, and on the cancel page the "Peru varaus"-button should also have a loading state now.
- Excluded from this PR/ticket: "Peru" buttons for the reservation process, as those will not result in any loading once the architecture is corrected - but will load instantly from cache.

## 🎫 Tickets

- TILA-2521

### DIFF
--- a/apps/ui/components/calendar/ReservationCalendarControls.tsx
+++ b/apps/ui/components/calendar/ReservationCalendarControls.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect, useState, useMemo } from "react";
-import { useTranslation } from "next-i18next";
+import { useTranslation, type TFunction } from "next-i18next";
 import styled from "styled-components";
 import {
   differenceInMinutes,
@@ -238,7 +238,7 @@ const ResetButton = styled(Button).attrs({
 
 const SelectButton = styled(Button)`
   order: 7;
-  ${truncatedText}
+  ${truncatedText};
 
   @media (min-width: ${breakpoints.m}) {
     display: none;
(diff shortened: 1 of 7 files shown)


apps/ui/components/calendar/ReservationCalendarControls.tsx
```diff
@@ -651,25 +652,25 @@ const ReservationCalendarControls = <T extends Record<string, unknown>>({
...
+  const TogglerLabelContent = areControlsVisible ? (
+    <div>&nbsp;</div>
+  ) : (
+    <>
+      <TogglerDate>{togglerLabel}</TogglerDate>
+      <TogglerPrice>
+        {t("reservationUnit:price")}: {price}
+      </TogglerPrice>
+    </>
+  );
```
joonatank, line 664: Generally bad. (yeah the above one is also bad).
Guess the linter rules don't block defining components inside components but they should.

Moving it is good, just not sure we want the downside.

Every time a component is rendered it creates a new component, destroys the old one, swaps the new into the virtual dom, which recreates the dom matching that component.

vincit-matu: Good point. Moved it outside the ReservationCalendarControls component.

apps/ui/components/calendar/ReservationCalendarControls.tsx
```diff
@@ -290,6 +290,23 @@ const StyledSelect = styled(Select<OptionType>)`
...
+  if (areControlsVisible) return <div>&nbsp;</div>;
+  return (
+    <>
+      <TogglerDate>{togglerLabel}</TogglerDate>
+      <TogglerPrice>
+        {t("reservationUnit:price")}: {price}
+      </TogglerPrice>
+    </>
+  );
+};
```
joonatank, line 308: JSX?

joonatank (approved):
LGTM, maybe change the `function` -> `JSX`

apps/ui/components/calendar/ReservationCalendarControls.tsx
vincit-matu: 👍

PR #1003, opened 2024-02-22
fix: remove todays reservastions from past list

joonatank:
## 🛠️ Changelog

- My reservations list shows yesterdays and older reservations (not todays).

## 🧪 Test plan

- Customer side my reservations list different combinations should never show today in past listing, and should always show today in upcoming even when the time of the reservation is already past.

## 🎫 Tickets

- TILA-####


vincit-matu (approved):
🚀

PR #1006, opened 2024-02-26
fix: disable reservation submit-button while the form isn't invalid

vincit-matu:
## 🛠️ Changelog

- Disabling the submit-button prohibits the user from clicking on it before they can advance (invalid form content won't let them continue) and ending up with a page without a submit button (due to isLoading being triggered and the spinner taking the button's place).

## 🧪 Test plan

- Go to make a reservation. Once in the reservation funnel (the part with the stepper, where you fill your info) you should no longer be able to press "Jatka" until the page form is valid (== until ticking necessary boxes in the case of the last step).

## 🎫 Tickets

- TILA-####


apps/ui/components/reservation/Step0.tsx
```diff
@@ -78,7 +78,7 @@ const Step0 = ({
 
   const {
     watch,
-    formState: { errors, isSubmitted },
+    formState: { errors, isSubmitted, isValid },
```
joonatank, line 81: Does `isSubmitting` work here instead of `isSubmitted`? Would fix issues with mutation failing and showing the loader still?

joonatank (changes requested):
The change here is fine, so the comments are more about the previous PR.

I'd check the EditPage also and confirm it doesn't need similar change.

I'd also check if we can change the `isSubmitted` -> `isSubmitting`, otherwise the loader is permanent after single submit try.

apps/ui/pages/reservation-unit/[...params].tsx
```diff
@@ -445,16 +445,20 @@ const ReservationUnitReservationWithReservationProp = ({
...
+  const onSubmitStep1 = (): Promise<void> | void => {
     confirmReservation({
       variables: {
         input: {
           pk: reservationPk ?? 0,
         },
       },
+    }).then(() => {
+      return Promise.resolve();
     });
```
Comment (joonatank) [inline apps/ui/pages/reservation-unit/[...params].tsx:462]: This is unnecessary

apps/ui/pages/reservation-unit/[...params].tsx
```diff
@@ -445,16 +445,20 @@ const ReservationUnitReservationWithReservationProp = ({
...
   };
 
-  const onSubmitStep1 = () => {
+  const onSubmitStep1 = (): Promise<void> | void => {
     confirmReservation({
```
Comment (joonatank) [inline apps/ui/pages/reservation-unit/[...params].tsx:454]: `return` missing

apps/ui/pages/reservation-unit/[...params].tsx
```diff
@@ -445,16 +445,20 @@ const ReservationUnitReservationWithReservationProp = ({
           reserveeLanguage: i18n.language,
         },
       },
+    }).then(() => {
```
Comment (joonatank) [inline apps/ui/pages/reservation-unit/[...params].tsx:448]: Missing `return` from `updateReservation`

joonatank (approved):
LGTM 👍 
As long as lint is ✅

PR #1007, opened 2024-02-27
fix: reservation form validity

vincit-matu:
## 🛠️ Changelog

- FIxes the validification issues which were still left:
  - now works for both a form that has reservee info included, and for a form that doesn't include it
  - the home city validation is fixed now, too
- The buttons with a loading state now have text in addition to the spinner (when isLoading is toggled)

## 🧪 Test plan

- Make a reservation for a reservation unit which **doesn't have** reservee information in the reservation funnel - you should be able to continue onwards when the form is valid (it used to be disabled, as it expected the reservee info for every form)
- Make a reservation for a reservation unit which **does** have required reservee information. There you should see, that "Jatka" only becomes enabled once you have also selected a home city from the dropdown. It used to become enabled even when no home city was selected.

## 🎫 Tickets

- TILA-2521


joonatank (approved):
LGTM 👍

PR #1008, opened 2024-02-27
fix: remove the 15min interval options from search duration filter

vincit-matu:
## 🛠️ Changelog

- Makes the search page duration option interval 30 minutes instead of the 15 it was previously.

## 🧪 Test plan

- Go to the search page and check out the options for duration. Try to make a search with some duration option selected to make sure the search didn't break.

## 🎫 Tickets

- TILA-????


joonatank (approved):
👍

PR #1009, opened 2024-02-28
Fix seconds causing reservation failure

joonatank:
## 🛠️ Changelog

- Fix: seconds sent through new reservation mutation causing it to always fail when using CalendarControls.
- Fix: permanent loading spinner when failing to create a reservation.

## 🧪 Test plan

- Try making a new reservation using the CalendarControls (below the reservation calendar), normally it shouldn't fail, but if it does the button should return back to non loading state.

## 🎫 Tickets

- TILA-####

### DIFF
--- a/apps/ui/pages/reservation-unit/[id].tsx
+++ b/apps/ui/pages/reservation-unit/[id].tsx
@@ -747,7 +747,7 @@ const ReservationUnit = ({
   });
 
   const createReservation = useCallback(
-    (res: ReservationProps): void => {
+    async (res: ReservationProps): Promise<void> => {
       setErrorMsg(null);
       setIsReserving(true);
       const { begin, end } = res;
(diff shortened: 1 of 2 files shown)


apps/ui/pages/reservation-unit/[id].tsx
```diff
@@ -765,11 +765,12 @@ const ReservationUnit = ({
         end,
       });
 
-      addReservation({
+      await addReservation({
```
Comment (vincit-matu) [inline apps/ui/pages/reservation-unit/[id].tsx:768]: I'd prefer using `then()`, for uniformity.

vincit-matu (approved):

apps/ui/pages/reservation-unit/[id].tsx
Comment (joonatank) [reply apps/ui/pages/reservation-unit/[id].tsx:768]: `async` `await` is prettier when there is a lot of them.
In this case it's not too bad, but in admin ui mutation code it's going to be `await` we are not gonna do
```
mut1.then((x) => {
  mut2.then((y) => {
   mut3.then(z) => {
   })
  })
})
```
So if consistency is important, `then` should be removed.

PR #1019, opened 2024-03-08
fix: set 255 maxLength for the metafields text inputs

vincit-matu:
## 🛠️ Changelog

- Adds maxLength to reservation input fields, as a quick fix

## 🧪 Test plan

- The fields in the ticket shouldn't allow longer values than 255 characters.
- The free-of-charge reason field length is not a problem, as it's not limited in the back end

## 🎫 Tickets

- TILA-3209


joonatank (approved):
👍

PR #1025, opened 2024-03-12
Fix reservation unit state

vincit-matu:
## 🛠️ Changelog

- Refactors the reservation unit page to use a single state (form)
- Single search now carries over the selected date, time, and duration to the reservation unit page (TODO)

## 🧪 Test plan

- Check that date, time and/or duration persists from the search page to the reservation unit page
- The reservation unit page controls (== input fields in quick reservation, the calendar, and the calendar controls) are in sync so changing one affects the others

## 🎫 Tickets

- TILA-2883, TILA-3042


apps/ui/pages/reservation-unit/[id].tsx
```diff
@@ -267,6 +284,11 @@ export const getServerSideProps = async (ctx: GetServerSidePropsContext) => {
         activeApplicationRounds,
         termsOfUse: { genericTerms: bookingTerms },
         isPostLogin: query?.isPostLogin === "true",
+        searchDuration: Number.isNaN(Number(duration))
+          ? null
+          : Number(duration),
+        searchDate: date ? `${date}` : null,
+        searchTime: time ? `${time}` : null,
```
Comment (joonatank) [inline apps/ui/pages/reservation-unit/[id].tsx:291]: `date` and `time` should be strings here why are they passed to template string?

apps/ui/pages/reservation-unit/[id].tsx
```diff
@@ -377,54 +413,143 @@ const ReservationUnit = ({
...
   const hash = router.asPath.split("#")[1];
 
-  const isClientATouchDevice = isTouchDevice();
+  const reservableTimeSpans = useMemo(
+    () => filterNonNullable(reservationUnit?.reservableTimeSpans),
+    [reservationUnit?.reservableTimeSpans]
+  );
+  const todaysTimeSpans = reservableTimeSpans.filter(
+    (span) => span.startDatetime && isToday(new Date(span.startDatetime))
+  );
```
Comment (joonatank) [inline apps/ui/pages/reservation-unit/[id].tsx:439]: Why `useMemo`? the second filter is slower than the one inside `useMemo`

apps/ui/pages/reservation-unit/[id].tsx
```diff
@@ -377,54 +413,143 @@ const ReservationUnit = ({
...
+  const searchUIDate = fromUIDate(searchDate ?? "");
+  // TODO: plug in query parameters
```
Comment (joonatank) [inline apps/ui/pages/reservation-unit/[id].tsx:441]: still?

apps/ui/pages/reservation-unit/[id].tsx
```diff
@@ -377,54 +413,143 @@ const ReservationUnit = ({
...
-      setInitialReservation(storedReservation);
-      removeStoredReservation();
-    } else if (hash === "calendar" && initialReservation) {
-      scrollToCalendar();
+      return [];
     }
+    return getDurationOptions(
+      minReservationDuration ?? undefined,
+      maxReservationDuration ?? undefined,
+      reservationStartInterval ?? 0,
```
Comment (joonatank) [inline apps/ui/pages/reservation-unit/[id].tsx:531]: Are `??` necessary? there is a `null` check already.

apps/ui/pages/reservation-unit/[id].tsx
```diff
@@ -377,54 +413,143 @@ const ReservationUnit = ({
...
+      durationValue
+    );
+  }, [
+    reservableTimeSpans,
+    reservationUnit?.reservationStartInterval,
+    now,
+    focusDate,
+    durationValue,
+    isSlotReservable,
+  ]);
```
Comment (joonatank) [inline apps/ui/pages/reservation-unit/[id].tsx:552]: This list look wrong. Why doesn't it match the params passed to `startingTimeOptions`

apps/ui/pages/reservation-unit/[id].tsx
```diff
@@ -518,26 +623,36 @@ const ReservationUnit = ({
...
-      const newReservation = getNewReservation({ start, end, reservationUnit });
-
       if (
-        !isSlotReservable(start, end, skipLengthCheck) ||
+        !isSlotReservable(
+          start,
+          end,
+          reservationUnit,
+          activeApplicationRounds
+        ) ||
```
Comment (joonatank) [inline apps/ui/pages/reservation-unit/[id].tsx:640]: This looks wrong. The same function is called with `skipLengthCheck` at the third position. ex. line 700

apps/ui/pages/reservation-unit/[id].tsx
```diff
@@ -953,8 +1070,10 @@ const ReservationUnit = ({
                 <div aria-hidden ref={calendarRef}>
                   <Calendar<ReservationType>
                     events={[...calendarEvents, ...eventBuffers]}
-                    begin={currentDate}
-                    onNavigate={(d: Date) => setFocusDate(d)}
+                    begin={focusDate}
+                    onNavigate={(d: Date) =>
+                      reservationForm.setValue("date", d.toISOString())
+                    }
```
Comment (joonatank) [inline apps/ui/pages/reservation-unit/[id].tsx:1076]: This looks wrong. isn't reservationForm dates in UI format?

apps/ui/pages/reservation-unit/[id].tsx
```diff
@@ -806,101 +930,94 @@ const ReservationUnit = ({
...
+          isLoading={reservationForm.formState.isSubmitting}
+          loadingText={t("reservationCalendar:makeReservationLoading")}
+          data-test="quick-reservation__button--submit"
+        >
+          {t("reservationCalendar:makeReservation")}
+        </SubmitButton>
+      }
+      returnUrl={getPostLoginUrl()}
+    />
+  );
```
Comment (joonatank) [inline apps/ui/pages/reservation-unit/[id].tsx:992]: should not create components inside components

apps/ui/components/calendar/ReservationCalendarControls.tsx
```diff
@@ -638,175 +377,154 @@ const ReservationCalendarControls = <T extends Record<string, unknown>>({
...
+              type="submit"
+              disabled={!focusSlot.isReservable}
+              isLoading={reservationForm.formState.isSubmitting}
+              loadingText={t("reservationCalendar:makeReservationLoading")}
+              data-test="reservation__button--submit"
+            >
+              {t("reservationCalendar:makeReservation")}
+            </SubmitButton>
           }
-        }}
```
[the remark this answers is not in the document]
vincit-matu: Agreed - I'd be tempted to "fix it all" in this PR - but with all the changes introduced here already it might be better to do a proper splitting-into-reasonably-sized-components themed branch/PR.

apps/ui/components/reservation/SubventionSuffix.tsx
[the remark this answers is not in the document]
vincit-matu: I think it shouldn't be needed by the code. It isn't relevant for anything other than maybe scrolling to the calendar.

apps/ui/components/common/ControlledSelect.tsx
```diff
@@ -3,14 +3,15 @@ import { Select } from "hds-react";
...
+import { getSelectedOption } from "@/modules/util";
 
 type Props = {
   name: string;
   label: string;
   control: ReturnType<typeof useForm>["control"];
-  required: boolean;
   options: OptionType[];
+  required?: boolean;
+  disabled?: boolean;
```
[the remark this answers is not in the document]
vincit-matu: Ahh..  it's in my other branch, where the calendar control starting time selection is disabled, if there are no available starting times for the selected day. This change should be on that branch, but it's here by mistake.

apps/ui/modules/reservation.ts
```diff
@@ -34,35 +33,41 @@ export const getDurationOptions = (
   if (!minReservationDuration || !maxReservationDuration || !intervalSeconds)
     return [];
 
-  const timeOptions = [];
+  const durationOptions: OptionType[] = [];
```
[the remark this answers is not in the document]
vincit-matu: Yeah, `OptionType` is problematic, but it's used _everywhere_. While I'm not a fan of coding more lines which have to be refactored later, I think making a quick-fix for it whenever encountered (just to avoid `OptionType`) is any better. Optimal solution would be to refactor away entire `OptionType` in one go.

apps/ui/components/Carousel.tsx
```diff
@@ -13,6 +13,7 @@ type Props = {
   cellSpacing?: number;
   wrapAround?: boolean;
   hideCenterControls?: boolean;
+  slideIndex?: number;
```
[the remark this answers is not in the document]
vincit-matu: It's also in the wrong branch :D This is used in the ui/ux-improvements branch to go to the correct "page"(== slide) of quick reservation slots, when a time is selected which isn't on the first page (otherwise the "quick slot selector" would show the first page, and thus "no selected slot" - even tho there would actually be a slot selected, but it's e.g. on the third page and hidden because of that).

apps/ui/modules/reservation.ts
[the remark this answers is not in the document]
joonatank: Yes but you changed the type of durationOptions so the refactor belongs here. Typescript checks break if there is old code left that assumes duration is `string`.

apps/ui/pages/reservation-unit/[id].tsx
Comment (vincit-matu) [reply apps/ui/pages/reservation-unit/[id].tsx]: That's absolutely correct - extracted it to a function component which is outside the parent component (`submitFragment`).

apps/ui/modules/reservation.ts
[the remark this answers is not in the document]
vincit-matu: 👍

apps/ui/components/Carousel.tsx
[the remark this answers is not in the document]
vincit-matu: Actually I take back my words about it being in the wrong branch. It's in use even here - in the eventuality that the query-params specify a time slot which is not on the first page/slide of the available quick reservation slots for the day - so the described feature it is warranted even here.

apps/ui/modules/reservationUnit.ts
```diff
@@ -337,34 +349,51 @@ export function isInTimeSpan(
 export function getPossibleTimesForDay(
   reservableTimeSpans: ReservationUnitByPkType["reservableTimeSpans"],
   reservationStartInterval: ReservationUnitByPkType["reservationStartInterval"],
-  date: Date
-): string[] {
+  date: Date,
+  reservationUnit: ReservationUnitByPkType,
+  activeApplicationRounds: RoundPeriod[],
+  durationValue: number
+): OptionType[] {
```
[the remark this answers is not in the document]
vincit-matu: Changed to `{ label: string; value: string; }`

apps/ui/pages/reservation-unit/[id].tsx
Comment (vincit-matu) [reply apps/ui/pages/reservation-unit/[id].tsx:290]: They can be `string`, `string[]` or `undefined` as they are taken directly from the query. Changed this to `.toString()` for (possibly?) clearer code.

apps/ui/pages/reservation-unit/[id].tsx
Comment (vincit-matu) [reply apps/ui/pages/reservation-unit/[id].tsx:467]: Ahhh - that's a brainfart. My thinking was that it should be on `todaysTimeSpans` (..so that we only filter today's time spans when `reservableTimeSpans` has changed).

apps/ui/pages/reservation-unit/[id].tsx
Comment (vincit-matu) [reply apps/ui/pages/reservation-unit/[id].tsx:469]: Well not exactly, but the form state should eventually only exist in the query parameters (i.e. the controlled inputs should also update the query params in the URL). However, this is not the optimal place to have a reminder about that entire refactoring need. Removed the comment.

apps/ui/pages/reservation-unit/[id].tsx
Comment (vincit-matu) [reply apps/ui/pages/reservation-unit/[id].tsx]: I ended up removing the `isSlotReservable()` function all together. It returns in practice the same thing as `isReservationReservable()`, so using the latter instead.

apps/ui/components/common/ControlledDateInput.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const { t } = useTranslation();
+
+  return (
+    <DateInput
+      id={id ?? `reservationDialog.${name}`}
+      label={label ?? t(`ReservationDialog.${name}`)}
+      minDate={minDate ?? new Date()}
+      maxDate={maxDate ?? addYears(new Date(), 2)}
+      initialMonth={initialMonth}
+      disableConfirmation={disableConfirmation ?? true}
```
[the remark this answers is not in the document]
vincit-matu: Good points.

joonatank (approved):
LGTM 👍

apps/ui/components/reservation-unit/ReservationInfoContainer.tsx
```diff
@@ -54,61 +49,65 @@ const ReservationInfoContainer = ({
...
-                    reservationUnit.reservationsMinDaysBefore,
-                }}
-                components={{ bold: <strong /> }}
-              />
-            )}
-          </p>
-        )}
+        {reservationUnitIsReservable &&
+          (reservationUnit.reservationsMaxDaysBefore != null ||
+            reservationUnit.reservationsMinDaysBefore != null) && (
```
joonatank, line 54: Is this some sonar lint thing? I don't like it. Creates extra nesting level for no reason in the JSX.

apps/ui/pages/reservation-unit/[id].tsx
```diff
@@ -371,61 +417,183 @@ any): JSX.Element => {
...
+    );
+  }, [reservationUnit, t]);
+  const minReservationDuration = reservationUnit.minReservationDuration
+    ? reservationUnit.minReservationDuration / 60
+    : 30;
+  const initialFieldValues = {
+    date:
+      searchDate && searchUIDate && isValidDate(searchUIDate)
+        ? searchDate
+        : toUIDate(new Date(todaysTimeSpans[0]?.startDatetime ?? "")),
```
Comment (joonatank) [inline apps/ui/pages/reservation-unit/[id].tsx:511]: Date format? here it's in UI.

apps/ui/components/calendar/ReservationCalendarControls.tsx
```diff
@@ -312,323 +269,79 @@ const TogglerLabelContent = ({
...
+  LoginAndSubmit,
+}: Props): JSX.Element => {
+  const { t } = useTranslation();
+  const { watch, handleSubmit } = reservationForm;
+  const { start, end } = focusSlot ?? {};
+  const formDate = watch("date");
+  const formDuration = watch("duration");
+  const date = new Date(formDate ?? "");
+  const dateValue = useMemo(() => new Date(formDate ?? ""), [formDate]);
+  const focusDate = useMemo(
```
joonatank, line 292: Date format? this is not valid if Date is in UI format (only ISO).

apps/ui/components/reservation/EditStep0.tsx
```diff
@@ -121,41 +134,149 @@ const EditStep0 = ({
...
   const isClientATouchDevice = isTouchDevice();
   const [shouldCalendarControlsBeVisible, setShouldCalendarControlsBeVisible] =
     useState(false);
+  const startDate = useMemo(
+    () => new Date(reservation.begin),
+    [reservation.begin]
+  );
+  const reservationForm = useForm<PendingReservationFormType>({
+    defaultValues: {
+      date: reservation.begin,
```
joonatank, line 153: Date format? here it's ISO string (directly from backend). 
Should be in UI format otherwise it's going to break HDS DateInput.

apps/ui/modules/reservationUnit.ts
```diff
@@ -223,54 +229,56 @@ export const getFuturePricing = (
...
+function formatPrice(
+  price: number,
+  trailingZeros: boolean,
+  toCurrency?: boolean
+) {
+  const currencyFormatter = trailingZeros ? "currencyWithDecimals" : "currency";
+  const floatFormatter = trailingZeros ? "twoDecimal" : "strippedDecimal";
+  const formatters = getFormatters("fi");
+  const formatter = formatters[toCurrency ? currencyFormatter : floatFormatter];
+  return parseFloat(price.toString()) ? formatter.format(price) : 0;
```
joonatank, line 241: ????

Why covert `number` -> `string` -> `number`?
Why return `string | number` union?

apps/ui/modules/reservationUnit.ts
```diff
@@ -223,54 +229,56 @@ export const getFuturePricing = (
     : futurePricings[0];
 };
 
+function formatPrice(
+  price: number,
+  trailingZeros: boolean,
+  toCurrency?: boolean
+) {
```
joonatank, line 236: Return value: is it supposed to be a `number` or `string`?
Now it's `string | number` which doesn't seem correct.

apps/ui/components/reservation-unit/QuickReservation.tsx
```diff
@@ -168,552 +136,175 @@ const SlotButton = styled.button`
...
-  );
-  const [duration, setDuration] = useState<OptionType | undefined>(
-    durationOptions.find((n) => n.value === "1:00") ?? durationOptions[0]
-  );
-  const [slot, setSlot] = useState<string | null>(null);
-  const [isReserving, setIsReserving] = useState(false);
+  const { t } = useTranslation();
+  const { setValue, watch, handleSubmit } = reservationForm;
+  const formDate = watch("date");
+  const dateValue = useMemo(() => new Date(formDate ?? ""), [formDate]);
```
joonatank, line 172: This seems off? is what's the Date format?

apps/ui/pages/reservation-unit/[id].tsx
Comment (vincit-matu) [reply apps/ui/pages/reservation-unit/[id].tsx:511]: It's supposed to be in UI format in the form/state.. `searchDate` is in UI format too..

apps/ui/components/calendar/ReservationCalendarControls.tsx
vincit-matu: Oh, that's right. Changed it to use the `fromUIDate()` helper function.

apps/ui/modules/reservationUnit.ts
vincit-matu: I have no idea - this is code that was already there. I might have relocated it but that whole `formatter` thing was already there (and seemed to work)..

apps/ui/components/reservation-unit/QuickReservation.tsx
vincit-matu: Yup, same thing as in ReservationCalendarControls

apps/ui/modules/reservationUnit.ts
joonatank: ok, it's awful, lets not touch it

PR #1055, opened 2024-03-20
Fix admin email validation

joonatank:
## 🛠️ Changelog

- Fix: admin-ui show email validation errors when making a reservation for a client (email is not mandatory, but if it is present it needs to be valid).
- Fix: admin-ui show backend error messages directly in the toast (for unknown cases), they are in clear Finnish now.
- Fix: admin-ui reservation list show previous results and loading indicator when filters or sort is changed, not empty state text.

## 🧪 Test plan

- Making a new reservation in the admin-ui should not fail with unknown or cryptic backend errors.
- There should not be an empty state text in the reservations list unless it's actually empty (too restrictive filtering etc.) it should show a table with previous data (grayed out) or current data (normal white) when filtering / sorting.

## 🎫 Tickets

- TILA-####


vincit-matu (approved):
LGTM

PR #1056, opened 2024-03-20
Fix application saving incorrect units

joonatank:
## 🛠️ Changelog

- fix: selected reservation units not on that application round got incorrectly included when saving the application.
- fix: not any reservation units selected validation error not correctly shown.
- fix: search showing incorrectly selected reservation units after changing application round filter.
- fix: application mobile layout delete and arrow buttons on top of each other.
- refactor: prefer gap over margin-top.

## 🧪 Test plan

- A new application should never include reservation units not on that application round.
- When selecting different round in the search the count at the bottom of the page should reflect only the reservation units on that round.
- Mobile delete and arrow buttons should be clickable and in the correct place.
- It should not be possible to save an application without any reservation units (or with just invalid reservation units) and it should show an missing field error to the user. 

## 🎫 Tickets

- TILA-####

### DIFF
--- a/apps/ui/components/application/ReservationUnitList.tsx
+++ b/apps/ui/components/application/ReservationUnitList.tsx
@@ -31,10 +31,13 @@ type Props = {
 
 const MainContainer = styled.div`
   margin-top: var(--spacing-l);
+  display: flex;
+  flex-direction: column;
+  gap: var(--spacing-l);
 `;
 
 const ButtonContainer = styled.div`
-  margin-top: var(--spacing-layout-m);
+  margin-top: var(--spacing-s);
 `;
 
 const Notification = styled(HDSNotification)`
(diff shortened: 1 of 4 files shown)


apps/ui/components/application/ReservationUnitList.tsx
```diff
@@ -53,7 +56,8 @@ const ReservationUnitList = ({
   const [showModal, setShowModal] = useState(false);
 
   const form = useFormContext<ApplicationFormValues>();
-  const { clearErrors, setError, watch, setValue } = form;
+  const { clearErrors, setError, watch, setValue, formState } = form;
+  const { errors } = formState;
```
vincit-matu, line 61: This could be just `const { clearErrors, setError, watch, setValue, formState: { errors } } = form;`, right? :)

joonatank: yeah was trying to avoid using 7 lines for it because of prettier

vincit-matu (approved):
Looks good 👍

apps/ui/components/application/ReservationUnitList.tsx
vincit-matu: Gotcha. 👍

PR #1075, opened 2024-03-25
fix: Image carousel look and accessibility

vincit-matu:
## 🛠️ Changelog

- adds custom bottom center controls for the image carousel on reservation-unit page
- aria-labels can be localised for the carousel center controls via the controlAriaLabel attribute
- the carousel modal closes only with esc-key, not any keyDown event
- new close-button and a bit of restyling for the image carousel modal

## 🧪 Test plan

- go to a reservation-unit page which has several images in it, and thus a carousel-element on the right
- inspect the bottom controls (circular small buttons - which are now bigger than before and themed to us)
- the aria-labels on the buttons should be dynamic and localized, so change language and check that the aria-label changes accordingly

## 🎫 Tickets

- TILA-3089


joonatank (approved):
LGTM 👍

PR #1131, opened 2024-04-22
fix: get csrf token before making graphql requests

joonatank:
## 🛠️ Changelog

- Fix: ui not working without being logged in after backend added csrf token checks to graphql endpoint.
- Graphql is only POST requests and Django wants the csrf token to be set before making a POST request. So if the cookie is not set we need to make separate dummy GET request on both SSR and Browser to fix that.
- Requires: https://github.com/City-of-Helsinki/tilavarauspalvelu-core/pull/1161
- Replaces: https://github.com/City-of-Helsinki/tilavarauspalvelu-ui/pull/1129

## 🧪 Test plan

- Deployment should always fail because of health checks if the SSR version doesn't work.
- Browser version can be checked from search page (if it says error, and there is 403 in the console it's broken).
- Manual with other tickets (if they are fine so is this).
- On client this affects primarily completely new sessions (destroy cookies for testing).

## 🎫 Tickets

- TILA-####

### DIFF
--- a/apps/ui/middleware.ts
+++ b/apps/ui/middleware.ts
@@ -10,11 +10,13 @@ import type { NextRequest } from "next/server";
 import { getSignInUrl } from "@/modules/const";
 import { env } from "@/env.mjs";
 
+const apiBaseUrl = env.TILAVARAUS_API_URL ?? "";
+
+/// should we redirect to the login page
 function redirectProtectedRoute(req: NextRequest) {
   // TODO check that the cookie is valid not just present
   const { cookies, headers } = req;
   const hasSession = cookies.has("sessionid");
-  const apiBaseUrl = env.TILAVARAUS_API_URL ?? "";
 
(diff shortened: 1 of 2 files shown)


apps/ui/middleware.ts
```diff
@@ -29,6 +31,40 @@ function redirectProtectedRoute(req: NextRequest) {
   return undefined;
 }
 
+/// are we missing a csrf token in cookies and should redirect to get one
+function redirectCsrfToken(req: NextRequest): URL | undefined {
+  // need to ignore all assets outside of html requests (which don't have an extension)
+  // so could we just check any request that doesn't have an extension?
+  if (
```
vergama, line 38: Would it work to limit redirects to paths matching apiBaseUrl + /graphql/ and /v1/ for REST?

vergama (approved):

PR #1138, opened 2024-04-25
Fix duration options not respecting min duration

joonatank:
## 🛠️ Changelog

- Fix: min reservation time check failed when generating duration options defaulting to reservation interval.
- Fix: reservation calendar not possible to drag to modify the selected time.
- Fix: webmanifest incorrectly redirected to core for csrf (adding errors in tests).
- Fix: clamp the duration query param (passed from search page) between min / max reservation duration.

## 🧪 Test plan

- Manual testing: Only normal user app changed. When making new reservations both calendar and quick reservation issues fixed.

## 🎫 Tickets

- TILA-####

### DIFF
--- a/apps/ui/pages/reservation-unit/[id].tsx
+++ b/apps/ui/pages/reservation-unit/[id].tsx
@@ -454,9 +454,13 @@ const ReservationUnit = ({
 
   const durationOptions = getDurationOptions(reservationUnit, t);
 
+  // technically these can be left empty (backend allows it)
   const minReservationDurationMinutes = reservationUnit.minReservationDuration
     ? reservationUnit.minReservationDuration / 60
     : 30;
+  const maxReservationDurationMinutes = reservationUnit.maxReservationDuration
+    ? reservationUnit.maxReservationDuration / 60
+    : Number.MAX_SAFE_INTEGER;
 
   const searchUIDate = fromUIDate(searchDate ?? "");
   // TODO should be the first reservable day (the reservableTimeSpans logic is too complex and needs refactoring)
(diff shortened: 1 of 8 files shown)


apps/ui/pages/reservation-unit/[id].tsx
```diff
@@ -476,7 +468,7 @@ const ReservationUnit = ({
       searchUIDate != null && isValidDate(searchUIDate)
         ? searchDate ?? ""
         : defaultDateString,
-    duration: searchDuration ?? minReservationDuration,
+    duration: searchDuration ?? minReservationDurationMinutes,
```
Comment (ranta) [inline apps/ui/pages/reservation-unit/[id].tsx:472]: Should this allow duration to be less than minimum duration?
e.g. User searched for 1 hour but minimum duration is 2 hours, then the duration would be 1h

Simple fix would be something like: `duration: Math.max(searchDuration ?? 0, minReservationDurationMinutes)`

apps/ui/pages/reservation-unit/[id].tsx
```diff
@@ -639,23 +642,32 @@ const ReservationUnit = ({
...
-        })
-      ) {
+      // the next check is going to systematically fail unless the times are at least minReservationDuration apart
+      const { minReservationDuration } = reservationUnit;
+      let newEnd = end;
+      if (minReservationDuration) {
+        if (differenceInSeconds(end, start) < minReservationDuration) {
+          newEnd = addSeconds(start, minReservationDuration);
+        }
+      }
```
Comment (ranta) [inline apps/ui/pages/reservation-unit/[id].tsx:652]: Slightly cleaner way:
```
let newEnd = end
if (minReservationDuration) {
    newEnd = Math.max(end, addSeconds(start, minReservationDuration))
}
```
(we just added new linter rules to backend for similar cases :D)

ranta (approved):
Didn't test locally but code looks fine :)
Would like it better if refactoring stuffs was done in a separate commit from fixes 😉

PR #1161, opened 2024-05-06
fix: my units calendar now shows the reservation name with prefix

vincit-matu:
## 🛠️ Changelog

- the staff/behalf prefix was not showing in the calendar entries on the my units page. now they are.

## 🧪 Test plan

- go to My Units, and make a reservation choosing the staff-option. The calendar entry should show "Sis. |" in front of the reservation name.

## 🎫 Tickets

- TILA-3171

### DIFF
--- a/apps/admin-ui/src/component/my-units/ReservationUnitCalendar.tsx
+++ b/apps/admin-ui/src/component/my-units/ReservationUnitCalendar.tsx
@@ -28,6 +28,7 @@ import {
 } from "common/src/helpers";
 import { RELATED_RESERVATION_STATES } from "common/src/const";
 import { ReservationUnitWithAffectingArgs } from "common/src/queries/fragments";
+import { TFunction } from "next-i18next";
 
 type Props = {
   begin: string;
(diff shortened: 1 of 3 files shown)


joonatank (commented):
The other calendar `ReservationUnitCalendar` doesn't need this?

Should we rather enforce `getReserveeName` to always take the TFunction (and fix the few instances that don't)?

apps/admin-ui/src/component/my-units/ReservationUnitCalendar.tsx
```diff
@@ -28,6 +28,7 @@ import {
 } from "common/src/helpers";
 import { RELATED_RESERVATION_STATES } from "common/src/const";
 import { ReservationUnitWithAffectingArgs } from "common/src/queries/fragments";
+import {t} from "i18next";
```
joonatank, line 31: This doesn't work correctly (needs to be prop drilled from `useTranstation`)

joonatank (approved):
👍

PR #1163, opened 2024-05-07
fix: check for correct user when displaying reservation

vincit-matu:
## 🛠️ Changelog

- Checks that the viewer is the owner of the reservation, so that reservation info can't be viewed by just anyone by typing in the url

## 🧪 Test plan

- Make a reservation with one user, go to its page and copy the url. Log in as another user and try to view that url.

## 🎫 Tickets

- TILA-3184

### DIFF
--- a/apps/ui/pages/reservations/[id]/index.tsx
+++ b/apps/ui/pages/reservations/[id]/index.tsx
@@ -1,5 +1,6 @@
 import React, { useMemo } from "react";
 import type { GetServerSidePropsContext } from "next";
+import Error from "next/error";
 import { serverSideTranslations } from "next-i18next/serverSideTranslations";
 import styled from "styled-components";
 import router from "next/router";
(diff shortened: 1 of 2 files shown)


apps/ui/pages/reservations/[id]/index.tsx
```diff
@@ -467,10 +468,13 @@ function Reservation({
   }, [reservation]);
 
   // NOTE should never end up here (SSR redirect to login)
-  if (!isAuthenticated) {
+  if (!isAuthenticated || reservation.user?.pk !== user?.pk) {
```
Comment (joonatank) [inline apps/ui/pages/reservations/[id]/index.tsx:471]: This is a left over (the next check does it)

joonatank (changes requested):

apps/ui/pages/reservations/[id]/index.tsx
Comment (vincit-matu) [reply apps/ui/pages/reservations/[id]/index.tsx]: 👍

joonatank (approved):
👍

PR #1166, opened 2024-05-07
fix: only relevant units are shown in search filter

vincit-matu:
## 🛠️ Changelog

- add the toggle to each search so only relevant units are shown in filter

## 🧪 Test plan

- Units with only seasonal and only direct reservations available aren't shown in single and seasonal searches respectively. In lieu of exhaustively checking every unit a simple test could be to see that the unit lists in the filter should be different on each of the search pages.

## 🎫 Tickets

- TILA-3247
- TILA-3248


joonatank (approved):
👍

PR #1179, opened 2024-05-15
fix: don't show error in unpaid notification if the time has run out

vincit-matu:
## 🛠️ Changelog

- shows the time until expiry dynamically (counting down the minutes in the front end, based on the value received in the query on page load)
- adds isLoading text to the cancel-button ("Canceling booking..."), so the button doesn't behave weirdly when in isLoading state (== have only the loading spinner as a label)

## 🧪 Test plan

- Make a reservation which requires payment, and stop the process without paying. 
- Open two duplicate tabs of the homepage 
- You should see the unpaid reservation notification banner in both of the duplicates.
- Cancel the reservation in one of them - should give you a notification that the reservation has been cancelled
- In the other tab (which still has the notification banner) click on "Cancel reservation" => this shouldn't result in any error messages, but simply reload the page.

## 🎫 Tickets

- TILA-3207

### DIFF
--- a/apps/ui/components/reservations/UnpaidReservationNotification.tsx
+++ b/apps/ui/components/reservations/UnpaidReservationNotification.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 import { useRouter } from "next/router";
 import { useTranslation } from "next-i18next";
 import styled from "styled-components";
@@ -19,6 +19,7 @@ import {
 import { getCheckoutUrl } from "@/modules/reservation";
 import { filterNonNullable } from "common/src/helpers";
 import { reservationUnitPrefix } from "@/modules/const";
+import { ApolloError } from "@apollo/client";
 
 const NotificationContent = styled.div`
   display: flex;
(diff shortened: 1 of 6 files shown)


apps/ui/components/reservations/UnpaidReservationNotification.tsx
```diff
@@ -44,6 +45,25 @@ const NotificationButtons = styled.div`
...
 
+function isNotFoundError(error: unknown): boolean {
+  if (error == null) {
+    return false;
+  }
+
+  if (error instanceof ApolloError) {
+    const { graphQLErrors } = error;
+    if (graphQLErrors.length > 0) {
+      // TODO separate validation errors: this is invalid form values (user error)
```
joonatank, line 56: remove comment?

apps/ui/components/reservations/UnpaidReservationNotification.tsx
```diff
@@ -57,20 +77,42 @@ function ReservationNotification({
...
-      : t("notification:waitingForPayment.payReservation");
+  const translateKey = isCreated
+    ? "notification:createdReservation"
+    : "notification:waitingForPayment";
+  const title = t(`${translateKey}.title`);
+  const submitButtonText = t(
+    `${translateKey}${isCreated ? ".continueReservation" : ".payReservation"}`
+  );
+  const text = t(`${translateKey}.body`, {
+    time: remainingMinutes,
```
joonatank, line 95: just to confirm, this doesn't affect the time of `Created`? since the value would be `0` or `null` here.

joonatank (approved):
check the few comments, otherwise lgtm 👍

apps/ui/components/reservations/UnpaidReservationNotification.tsx
vincit-matu: No it shouldn't. The `Created` variant shows 20 minutes as it's a different translation key. (`waitingForPayment` vs `createdReservation`)

apps/ui/components/reservations/UnpaidReservationNotification.tsx
vincit-matu: Removed

PR #1223, opened 2024-06-06
fix: missing free of charge reason in summary

joonatank:
## 🛠️ Changelog

- Fix: Reservation summary page (step2) didn't show the reason user added when applying for a free of charge (only that the user applied for it).
- The field was missing in GQL query, and it's not caught with the type checker because metafields are dynamically constructed from strings and retrieved from the Record, not directly addressed from the query object.

## 🧪 Test plan

- Manual testing: summary page should now include the reason user inputted (and should not if none was applied for).

## 🎫 Tickets

- TILA-####


vincit-matu (approved):
Good catch! 👍

PR #1416, opened 2024-09-17
add: docker build pipeline for ui

joonatank:
## 🛠️ Changelog

- Add: docker build action

## 🧪 Test plan

- Automation: deployment.

## 🎫 Tickets

- [TILA-3513](https://helsinkisolutionoffice.atlassian.net/browse/TILA-3513)

[TILA-3513]: https://helsinkisolutionoffice.atlassian.net/browse/TILA-3513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Dockerfile.final
+++ b/Dockerfile.final
@@ -1,0 +1,3 @@
+ARG TAG=main
+ARG APP=ui
+FROM ghcr.io/city-of-helsinki/tilavarauspalvelu-ui-$APP:$TAG AS $APP
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -1,0 +1,83 @@
+# Generic workflow to build a frontend application and push it to the GitHub Container Registry.
+name: "Build docker"
+
+on:
+  workflow_call:
+    inputs:
+      # Docker build arg / package name suffix: valid values are "admin-ui" or "ui"
+      app-name:
+        required: true
+        type: string
+      # Root url for the application relative to hostname
+      base-url:
+        required: true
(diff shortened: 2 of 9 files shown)


Dockerfile.final
```diff
@@ -0,0 +1,3 @@
+ARG TAG=main
+ARG APP=ui
+FROM ghcr.io/city-of-helsinki/tilavarauspalvelu-ui-$APP:$TAG AS $APP
```
matti-lamppu, line 3: Should this be `tilavarauspalvelu-$APP` instead?

matti-lamppu (changes requested):
`Dockerfile.final` might be wrong.

Dockerfile.final
joonatank: No because of the way the name is created
`echo ${{ env.REGISTRY }}/${{ env.REPO_NAME }}-${{ env.BUILD_APP }} | tr '[:upper:]' '[:lower:]'`

Otherwise have to do some regex / string magic to strip the `-ui` part of the repo name which I think is more fragile than repeating. 

The package names always include `tilavarauspalvelu-ui` (as per repo name) and the app name and we end up with `tilavarauspalvelu-ui-ui` and `tilavarauspalvelu-ui-admin-ui`. I'd rather rename the apps than not use the repo name as part of the package name.

matti-lamppu (approved):

PR #1398, opened 2024-09-10
Add: universal Card component

vincit-matu:
## 🛠️ Changelog
[//]: # "Describe the changes in this pull request here."

Adds a `Card` component to common components, which displays a plain card with `heading` and `text`.
It can also display:
  - an image (if given an `imageSrc). The image uses `heading`as `alt` text by default, but `imageAlt` allows the use of a any value.
  - `infos` (a `value` with an optional `icon` and `testId`)
  - `tags` ((== `JSX.Element`s, like `StatusLabel`s and `Tag`s)
  - `buttons` (== `JSX.Element`s, like `Button`s, `ButtonLikeLink`s etc)
  - child elements, for the moment with the same positioning as `infos`, as our use-cases are mutually exclusive between the two
- if given a `link`, it is used with the image and heading text
- any child elements will be displayed after the main `text`

Use `margin` to apply one on the entire card, `backgroundColor` for a custom background color, and `testId`/`headingTestId`/`textTestId` to provide `data-test-id`s for each element respectively. To use `data-test-id`s for `tags` and `buttons` they should be included in JSX.Elements you provide)

The replaced cards:
 - Single reservation: search result card and my reservations listing cards
 - Recurring reservations: application round card and my applications cards
 - Reservation unit page: related units cards, on the bottom of the screen
 - Admin pages reservation unit card, on a unit's page

## 🧪 Test plan
[//]: # "Help your fellow reviewer and write a short description of what's the fastest way to test your changes."

- Manual testing of the different cards in different screen sizes.

## 🎫 Tickets
[//]: # "This pull request resolves all or part of the following ticket(s)"

- TILA-3522

PR #1427, opened 2024-09-23
fix: tag and status label borders

vincit-matu:
## 🛠️ Changelog

- Adds `Tag.tsx` as a common component. It is a HDS Tag with the same custom types and coloring as `StatusLabel`
- Moves code which is shared by both Tag and StatusLabel (e.g. the color functions based on type) from StatusLabel.tsx to tags.tsx
- renames StatusLabel property `dataTestId` to `testId`, just for consistency

## 🧪 Test plan

- Check that StatusLabels and Tags have a border, like specified in the designs
- Check that tags still otherwise work like they did

## 🎫 Tickets

- TILA-####

### DIFF
--- a/packages/common/src/components/Tag.tsx
+++ b/packages/common/src/components/Tag.tsx
@@ -1,0 +1,44 @@
+import React from "react";
+import {
+  getStatusBackgroundColor,
+  getStatusBorderColor,
+  StatusLabelType,
+  StyledTag,
+} from "../tags";
+import styled from "styled-components";
+
+type TagPropsType = {
+  ariaLabel: string;
+  type: StatusLabelType;
+  onClick?: () => void;
+  id?: string;
+  children: React.ReactNode;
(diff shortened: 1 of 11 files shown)


packages/common/src/components/Tag.tsx
```diff
@@ -0,0 +1,44 @@
...
+} from "../tags";
+import styled from "styled-components";
+
+type TagPropsType = {
+  ariaLabel: string;
+  type: StatusLabelType;
+  onClick?: () => void;
+  id?: string;
+  children: React.ReactNode;
+} & Pick<React.HTMLAttributes<HTMLDivElement>, "role">;
```
joonatank, line 16: why? `role` is not used.

joonatank (approved):
ok, but check the comment

vincit-matu:
These  were already implemented in #1398 - so closing the PR

PR #1503, opened 2024-11-12
Add deactivated account error-page

vincit-matu:
## 🛠️ Changelog

- Adds a page to show if the client account has been deactivated

## 🧪 Test plan

- Go to https://tilavaraus.test.hel.ninja/deactivated-account and compare it to the [design](https://www.figma.com/design/s987FF5VRRUNhDBvZJFPs9/HEL-Varaamo?node-id=5241-491941&t=llCEpG8dnjac9gOr-4).

## 🎫 Tickets

- TILA-3629


apps/admin-ui/src/App.tsx
```diff
@@ -151,6 +152,15 @@ function ClientApp({
               UserPermissionChoice.CanManageNotifications
             )}
           />
+          <Route
+            path="/deactivated-account"
+            element={
+              <DeactivatedAccount
+                feedbackUrl={`${feedbackUrl}foo`}
```
joonatank, line 159: what's the `foo` for?

packages/common/src/components/DeactivatedAccount.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { IconArrowRight } from "hds-react";
+import { fontBold, fontRegular } from "../common/typography";
+
+const ErrorContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  max-width: var(--container-width-xl);
+  margin: 0 auto;
+  gap: 42px;
+  padding-block: 42px;
```
joonatank, line 14: why the pixel sizes?

packages/common/src/components/DeactivatedAccount.tsx
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+`;
+
+const TextContent = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 42px;
+  justify-content: center;
+  @media (min-width: 768px) {
```
joonatank, line 43: why use pixel size instead of `breakpoint.m`?

packages/common/src/components/DeactivatedAccount.tsx
```diff
@@ -0,0 +1,116 @@
...
+      max-width: 350px;
+      flex-grow: 1;
+    }
+  }
+`;
+
+const TextContent = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 42px;
```
joonatank, line 41: why pixels? and why such large one? why not just use `--spacing-m` like everywhere else?

packages/common/src/components/DeactivatedAccount.tsx
```diff
@@ -0,0 +1,116 @@
...
+  justify-content: center;
+  @media (min-width: 768px) {
+    order: -1;
+  }
+`;
+
+const Heading = styled.h1`
+  margin: 0;
+  line-height: 1;
+  font-size: 40px;
```
joonatank, line 51: why? just use typography `H1`

packages/common/src/components/DeactivatedAccount.tsx
```diff
@@ -0,0 +1,116 @@
...
+  ${fontRegular}
```
joonatank, line 52: nope. we don't modify the fonts per page for `h1`

inherit from `H1` and it's what it is on other pages.

packages/common/src/components/DeactivatedAccount.tsx
```diff
@@ -0,0 +1,116 @@
...
+`;
+
+const Body = styled.p`
+  margin: 0;
+`;
+
+const Email = styled.a`
+  margin: 8px 0 0;
```
joonatank, line 60: why pixel margins?

and why margins at all? is there a styling issue for `<p><a>`? that seems a lot more fundamental problem than just chucking random `8px` top margins to some links

if the design calls for it. then either split it into separate `<p>`, use `line-height`, or the design is wrong

packages/common/src/components/DeactivatedAccount.tsx
```diff
@@ -0,0 +1,116 @@
...
+  ${fontBold}
+`;
+
+// NOTE: This is a copy of the Footer component from the common package
```
joonatank, line 64: why copy it? we are in common?

packages/common/src/components/DeactivatedAccount.tsx
```diff
@@ -0,0 +1,116 @@
...
+  imgSrc,
+}: {
+  feedbackUrl: string;
+  imgSrc: string;
+}) => {
+  console.log(feedbackUrl);
+  const { t, i18n } = useTranslation();
+  return (
+    <ErrorContainer>
+      <img
```
joonatank, line 89: `next/image` still not working after you fixed the img source path?

packages/common/src/components/DeactivatedAccount.tsx
```diff
@@ -0,0 +1,116 @@
...
+import styled from "styled-components";
+import IconButton from "./IconButton";
+import { IconArrowRight } from "hds-react";
+import { fontBold, fontRegular } from "../common/typography";
+
+const ErrorContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  max-width: var(--container-width-xl);
+  margin: 0 auto;
```
joonatank, line 12: This whole thing has to be redone after layouts have been merged.
So much copy pasta `margin`, `padding`, `max-width` code.

But if this is merged first we can fixit then.

packages/common/src/components/DeactivatedAccount.tsx
```diff
@@ -0,0 +1,116 @@
...
+    flex-direction: row;
+    padding-inline: var(--spacing-m);
+    --spacing-hz: var(--spacing-m);
+    img {
+      display: flex;
+      margin: 0;
+      margin-inline: auto;
+      width: auto;
+      max-height: 418px;
+      max-width: 350px;
```
joonatank, line 32: not a fan of these magic numbers. where are they from? (except from Figma for this particular page)

packages/common/src/components/DeactivatedAccount.tsx
```diff
@@ -0,0 +1,116 @@
...
+  display: flex;
+  flex-direction: column;
+  max-width: var(--container-width-xl);
+  margin: 0 auto;
+  gap: 42px;
+  padding-block: 42px;
+  padding-inline: var(--spacing-s);
+  --spacing-hz: var(--spacing-s);
+  img {
+    width: 185px;
```
joonatank, line 18: magic values, why?

packages/common/src/components/DeactivatedAccount.tsx
```diff
@@ -0,0 +1,116 @@
...
+    height: auto;
+    margin: 0 auto;
+  }
+  @media (min-width: 768px) {
```
joonatank, line 22: use var

joonatank (changes requested):

packages/common/src/components/DeactivatedAccount.tsx
vincit-matu: All references to breakpoints switched to actually reference `breakpoints`. 👍

packages/common/src/components/DeactivatedAccount.tsx
vincit-matu: All "magic numbers" switched to using --spacing-variables (apart from the image sizes, which are indeed a bit magic).

packages/common/src/components/DeactivatedAccount.tsx
vincit-matu: This margin doesn't have any special reason to be in pixels. It is actually just left behind code, now removed as unnecessary. It didn't have any visible effect since it was block margin on an inline element (which is probably why I didn't notice it was left behind).

packages/common/src/components/DeactivatedAccount.tsx
vincit-matu: That's a `NOTE` left behind from before it was a common component. Removed.

packages/common/src/components/DeactivatedAccount.tsx
vincit-matu: Yeah, now it seems to work on both sides => using `next/image` instead of `img`.

packages/common/src/components/DeactivatedAccount.tsx
vincit-matu: I tried removing as much of the extraneous css-stuff as possible after rebasing it on the current main, so it should look a bit less crowded in that section. Now all of the remaining rules seem logical/justified, to me at least.

packages/common/src/components/DeactivatedAccount.tsx
vincit-matu: As many as possible of the magic values removed, all of the remaining ones pertain to the image dimensions.

packages/common/src/components/DeactivatedAccount.tsx
vincit-matu: Yeah, they're "magic" numbers - and are based on the image dimensions in Figma.

packages/common/src/components/DeactivatedAccount.tsx
vincit-matu: Also, switched the image dimensions into their own `const`s - to lessen the "magic" component :D

PR #1506, opened 2024-11-13
Fix observed bugs in applications

vincit-matu:
## 🛠️ Changelog

- fixes status labels styling, categories and translations according to the reported bugs
- the box about receiving a notification when the application is handled is no longer shown when viewing applications that are already handled
- StatusLabel font explicitly weight set to regular, so it doesn't inherit bold from event group heading

## 🧪 Test plan

- Check on my applications page that the status labels are correct (i.e. they correspond to the heading they're under)
- View an application from the second group, "Käsittelyssä", and note the blue notification box on the bottom labelled "Käsittely"
- That box should not be there when viewing applications that have the "Käsitelty" or "Luonnos" statuses
- The StatusLabel components located next to application event group headings have regular font weight

## 🎫 Tickets

- TILA-3680

### DIFF
--- a/apps/ui/components/application/ViewApplication.tsx
+++ b/apps/ui/components/application/ViewApplication.tsx
@@ -1,10 +1,11 @@
 import React from "react";
 import { Checkbox } from "hds-react";
 import { useTranslation } from "next-i18next";
-import type {
-  ApplicationQuery,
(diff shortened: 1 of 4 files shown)


joonatank (approved):
👍

PR #1553, opened 2024-11-28
Fix dropdown options for client

vincit-matu:
## 🛠️ Changelog

- Uses the non-paged endpoints for units and equipment filter options
- Fixes the sorting of options on the search page filters in client view as follows:
  - *purpose:* by rank
  - *unit:* by name
  - *equipments:* by equipments category rank
  - *reservation type:* by rank
- Sorts the reservation purpose options by rank, when making a reservation on client side

## 🧪 Test plan

- Check corresponding name/rank value from Django, and see that the options in respective select-inputs are in the same order on the search page
- Do the same for reservation purposes

## 🎫 Tickets

- TILA-3541
- TILA-3553

### DIFF
--- a/apps/ui/hooks/useOptions.ts
+++ b/apps/ui/hooks/useOptions.ts
@@ -1,6 +1,12 @@
 import { gql } from "@apollo/client";
 import { useTranslation } from "next-i18next";
-import { type Maybe, useOptionsQuery, type OptionsQuery } from "@gql/gql-types";
+import {
+  type Maybe,
+  type OptionsQuery,
+  ReservationPurposeOrderingChoices,
+  ReservationUnitTypeOrderingChoices,
+  useOptionsQuery,
+} from "@gql/gql-types";
 import { participantCountOptions } from "@/modules/const";
 import { filterNonNullable, getLocalizationLang } from "common/src/helpers";
 import { getParameterLabel } from "@/modules/util";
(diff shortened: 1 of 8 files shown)


apps/ui/modules/search.ts
```diff
@@ -310,21 +317,7 @@ export async function getSearchOptions(
...
-
-  const { data: unitData } = await apolloClient.query<
-    SearchFormParamsUnitQuery,
-    SearchFormParamsUnitQueryVariables
-  >({
-    query: SearchFormParamsUnitDocument,
-    variables: {
-      publishedReservationUnits: true,
-      ...(page === "direct" ? { onlyDirectBookable: true } : {}),
-      ...(page === "seasonal" ? { onlySeasonalBookable: true } : {}),
```
joonatank, line 322: are these filters not present in the new endpoint? they should be otherwise this introduces a regression.

joonatank (changes requested):
missing unit query params (seasonal / single) need to be checked

apps/ui/modules/search.ts
vincit-matu: No, they don't seem to be present in `unitsAll`. I brought it up in #tilanvaraus-hyväksymistestaus - basically asking if it should/could be added to this query, too, to preserve the functionality which depends on them.

apps/ui/hooks/useOptions.ts
```diff
@@ -124,7 +151,11 @@ function sortAgeGroups(ageGroups: AgeGroup[]): AgeGroup[] {
 export function useOptions() {
   const { i18n } = useTranslation();
 
-  const { data, loading: isLoading } = useOptionsQuery();
+  const { data, loading: isLoading } = useOptionsQuery({
+    variables: {
+      reservationPurposesOrderBy: ReservationPurposeOrderingChoices.RankAsc,
```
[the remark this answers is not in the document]
vincit-matu: The hook returns `options` and `params` which are in turn formed from `ageGroups`, `cities`, `purposes` and `reservationUnitTypes`. And yeah, like you say `ageGroups` is taken care of in the front. Meanwhile `cities` are shown in the order the backend provides them == `cities` doesn't have an `orderBy` argument. That leaves only the last two to be ordered here. Code changed to reflect this reasoning.

joonatank:
can't be merged before the query params are added to backend and the query is changed

apps/ui/modules/search.ts
vincit-matu: Added this back, after changes to the backend

joonatank (approved):
👍

PR #1578, opened 2024-12-15
Application flow fixes

vincit-matu:
## 🛠️ Changelog

- Addresses the issues listed on the application flow ticket (TILA-3560), apart from ones to do with validating that the specific timespans correspond to the given amount and duration in the application
- Fixes page2 of the application flow in respect to the application sections, and selected reservation units

## 🧪 Test plan

- Check out the application flow, comparing it to [the design in figma](https://www.figma.com/design/s987FF5VRRUNhDBvZJFPs9/HEL-Varaamo?node-id=4318-369406&t=fHO7a0Yu1DvlTBH6-1).
- Make an application with more than one sections and note, that you can now enter the primary/secondary timeslots and the form should all-in-all work as expected when filling in details for more than one section in one application. Using sections with different selections of spaces makes this easier to see at a glance.

## 🎫 Tickets

- TILA-3560


apps/ui/components/application/ApplicationPage.tsx
```diff
@@ -91,35 +97,64 @@ export function ApplicationPageWrapper({
...
+      push(`${getApplicationPath(application?.pk)}preview`);
+    } else {
+      push(`${getApplicationPath(application?.pk)}page${s}`);
+    }
   };
 
   const title = t(`${translationKeyPrefix}.heading`);
   const subTitle =
     headContent || overrideText || t(`${translationKeyPrefix}.text`);
+  const selectedStep = asPath.charAt(asPath.length - 1);
```
joonatank, line 140: Complete word comparisons if we have to convert strings to numbers.
If somebody changes the slug name this will break, without nothing indicating that you should have changed the comparison also.

apps/ui/components/application/ApplicationPage.tsx
```diff
@@ -91,35 +97,64 @@ export function ApplicationPageWrapper({
...
       <BreadcrumbWrapper route={["/applications", "application"]} />
       <div>
         <H1 $noMargin>{title}</H1>
         <p>{subTitle}</p>
       </div>
+      {hideStepper ? null : (
+        <StyledStepper
+          language={i18n.language}
+          steps={steps}
+          selectedStep={selectedStep === "w" ? 3 : Number(selectedStep) - 1} //calculateCompletedStep(application)}
```
joonatank, line 153: Use a `slug` => `step: Number` map where `slug` is the last word of the url. either JS object or a function with switch statement.

also Number conversion is unsafe and should be done in the function body not inside JSX (with NaN checks) or use `toNumber` and check for nulls.

and the comment can be removed

apps/ui/components/application/ApplicationPage.tsx
```diff
@@ -91,35 +97,64 @@ export function ApplicationPageWrapper({
...
+          ? StepState.disabled
+          : StepState.completed,
+  }));
+
+  const handleStepClick = (e: React.MouseEvent<HTMLButtonElement>) => {
+    const target = e.currentTarget;
+    const s =
+      Number(
+        target.getAttribute("data-testid")?.replace("hds-stepper-step-", "")
+      ) + 1;
```
joonatank, line 126: There isn't a proper way to pass data in the step handler?

and do NaN checks (or use `toNumber`) helper.

apps/ui/components/application/reservation-unit-card.tsx
```diff
@@ -189,66 +221,51 @@ export function ReservationUnitCard({
...
-            <button
-              className="button-reset"
-              disabled={first}
-              type="button"
-              aria-label={t("reservationUnitList:buttonUp")}
-              onClick={() => onMoveUp(reservationUnit)}
+        <OrderButtonContainer>
+          <DeleteContainer>
+            <DeleteButton
+              iconRight={undefined}
```
joonatank, line 243: why? it default's to `undefined`

apps/ui/components/application/reservation-unit-card.tsx
```diff
@@ -189,66 +221,51 @@ export function ReservationUnitCard({
...
+          </DeleteContainer>
+          <UpButton
+            iconLeft={<IconArrowUp aria-hidden />}
+            onClick={() => onMoveUp(reservationUnit)}
+            disabled={first}
+          >
+            {t("reservationUnitList:buttonUp")}
+          </UpButton>
+          <DownButton
+            iconLeft={<IconArrowDown aria-hidden />}
```
joonatank, line 261: `aria-hidden="true"`

apps/ui/components/application/reservation-unit-card.tsx
```diff
@@ -189,66 +221,51 @@ export function ReservationUnitCard({
...
-              <IconArrowDown aria-hidden size="m" />
-            </button>
-          </Circle>
-        </ArrowContainer>
-      </CardButtonContainer>
+              {t("reservationUnitList:buttonRemove")}
+            </DeleteButton>
+          </DeleteContainer>
+          <UpButton
+            iconLeft={<IconArrowUp aria-hidden />}
```
joonatank, line 254: `aria-hidden="true"`

joonatank (changes requested):
the stepper comparison is questionable

apps/ui/components/application/ApplicationPage.tsx
vincit-matu: The HDS Stepper passes the stepIndex as second argument, so it's easier to just use that

apps/ui/components/application/reservation-unit-card.tsx
vincit-matu: TS check gives lint error if `iconRight` is not defined at all - which is weird, as even in the error description states that `iconRIght` should be optional for the HDS Button (the type is `iconRight?: React.ReactNode`).

...but that's why it's `undefined`.

vincit-matu:
Addressed the issues raised by @joonatank

apps/ui/components/application/reservation-unit-card.tsx
```diff
@@ -28,135 +24,171 @@ type Props = {
...
+    ${fontBold}
+    &:hover,
+    &:focus:hover {
+      background-color: var(--color-black-5);
+    }
 
-const UnitName = styled.div``;
+    &:focus {
+      background-color: var(--color-white);
+    }
```
joonatank, line 139: this seems super random. if overriding HDS colors use the vars if creating actual hover / focus styles import `focusStyles` from `cssFragments.ts`

apps/ui/components/application/reservation-unit-card.tsx
```diff
@@ -28,135 +24,171 @@ type Props = {
...
-  @media (min-width: ${breakpoints.m}) {
-    place-self: flex-end;
+      &:hover,
+      &:focus:hover {
+        background-color: var(--color-black-10);
+      }
+
+      &:focus {
+        background-color: var(--color-black-5);
+      }
```
joonatank, line 159: doubling up on focus / hover

apps/ui/components/application/reservation-unit-card.tsx
```diff
@@ -28,135 +24,171 @@ type Props = {
...
   }
 `;
 
-const Name = styled.div`
-  ${fontBold};
-  font-size: var(--fontsize-heading-s);
-  line-height: var(--lineheight-xl);
-  margin-bottom: var(--spacing-3-xs);
-`;
+const OrderButton = styled(Button)`
```
joonatank, line 124: Don't split `:focus` / `:hover` / `:disabled` styles by `@media`. One definition for each if necessary. It's a HDS Button so they should not be overridden like this any way (CSS overrides don't typically work for HDS components), use CSS vars.

joonatank (changes requested):
CSS is super weird and verbose for the Card. the amount of `!important` seems that something is wrong either use HDS CSS vars or don't use HDS Buttons as base.

If special buttons start requiring 48 lines of CSS maybe we should talk to the Designer and make them into components.

apps/ui/components/application/reservation-unit-card.tsx
```diff
@@ -28,135 +24,171 @@ type Props = {
...
+    flex-direction: row;
+  }
+`;
 
 const PreCardLabel = styled(H6).attrs({ as: "h3" })`
   margin-bottom: 0;
   margin-top: 0;
+  @media (min-width: ${breakpoints.m}) {
+    display: flex;
+    align-items: center;
```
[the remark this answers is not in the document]
vincit-matu: Since the h3 contains the span ("Vaihtoehto") which isn't shown in desktop, they line up correctly as it is now. If we use flex in mobile, we'd need to wrap the order number in a separate span, and style it separately, resulting in even more CSS..
So I think it's a better trade-off the way it is.

apps/ui/components/application/reservation-unit-card.tsx
```diff
@@ -28,135 +24,171 @@ type Props = {
...
-  font-size: var(--fontsize-heading-s);
-  line-height: var(--lineheight-xl);
-  margin-bottom: var(--spacing-3-xs);
-`;
+const OrderButton = styled(Button)`
+  && {
+    position: relative;
+    z-index: 2;
+    color: var(--color-black-90) !important;
+    background-color: var(--color-white);
```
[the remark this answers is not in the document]
vincit-matu: Nope, it's blue by default. But changed it to use a css-variable, as per the other comments.

apps/ui/components/application/reservation-unit-card.tsx
```diff
@@ -28,135 +24,171 @@ type Props = {
...
-  ${fontBold};
-  font-size: var(--fontsize-heading-s);
-  line-height: var(--lineheight-xl);
-  margin-bottom: var(--spacing-3-xs);
-`;
+const OrderButton = styled(Button)`
+  && {
+    position: relative;
+    z-index: 2;
+    color: var(--color-black-90) !important;
```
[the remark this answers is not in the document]
vincit-matu: Yeah, that's right - and it's the reason for all these `!important`s. Will change.

apps/ui/components/application/reservation-unit-card.tsx
```diff
@@ -28,135 +24,171 @@ type Props = {
...
+    background-color: var(--color-white);
+    border-color: var(--color-black-90) !important;
```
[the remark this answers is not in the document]
vincit-matu: Changed to use css-variable

apps/ui/components/application/reservation-unit-card.tsx
vincit-matu: Might _seem_ random, but there's a method to the madness. We need to override the blueness which HDS-buttons use as default. But changed it to use css-variables.

apps/ui/components/application/reservation-unit-card.tsx
```diff
@@ -28,135 +24,171 @@ type Props = {
...
+  margin-left: -2px !important;
+  margin-right: auto !important;
+  @media (min-width: ${breakpoints.m}) {
+    margin-left: 0 !important;
+    margin-right: 0 !important;
+    grid-column: 2;
+    grid-row: 2;
+    &&:disabled {
+      border-right: 0 !important;
     }
```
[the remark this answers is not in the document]
vincit-matu: Yeah, it needs more specificity - but `&&` seems to do the trick!

apps/ui/components/application/reservation-unit-card.tsx
vincit-matu: Changed these to use the HDS css-variables, and moved them to the button (instead of to the `:focus`/ `:hover`).

PR #1646, opened 2025-02-02
add: "Cancelled" status label to recurring reservation lists

vincit-matu:
## 🛠️ Changelog

- Adds a "Cancelled" `StatusLabel`to recurring reservation listings, when the user has cancelled a reservation (vs e.g. being denied.
- Changes the overlapping reservations StatusLabel text to simply "Ei saatavilla" 

## 🧪 Test plan

- As a user with a recurring reservation series, cancel a reservation -> it should have a `StatusLabel` "Cancelled"
- As admin go see the same recurring reservation, and open the "Toistokerrat" accordion. It should also have a `StatusLabel` instead of the _edit-_, _show in calendar-_ and _deny_-buttons

## 🎫 Tickets

- TILA-3758

### DIFF
--- a/apps/admin-ui/src/component/ReservationListButton.tsx
+++ b/apps/admin-ui/src/component/ReservationListButton.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import {
   Button,
+  ButtonPresetTheme,
   ButtonSize,
   ButtonVariant,
   IconArrowUndo,
@@ -17,16 +18,17 @@ export function ReservationListButton({
   type,
   callback,
   t,
-}: {
+}: Readonly<{
   type: "remove" | "deny" | "restore" | "change" | "show";
   callback: () => void;
   // Pass the TFunc because the amount of buttons change and hooks break
   t: TFunction;
-}) {
+}>) {
   const btnCommon = {
     variant: ButtonVariant.Supplementary,
     onClick: callback,
     size: ButtonSize.Small,
+    theme: ButtonPresetTheme.Black,
   } as const;
 
   switch (type) {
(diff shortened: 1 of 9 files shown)


joonatank (changes requested):
`aria-hidden` on icons is unnecessary and creates confusion when you actually need the label to be visible since HDS defaults all icons to `aria-hidden=true`

apps/admin-ui/src/component/ReservationsList.tsx
```diff
@@ -88,18 +85,28 @@ function getStatus(x: NewReservationListItem) {
     return {
       isError: true,
       msg: "overlapping",
+      icon: <IconCross aria-hidden />,
```
joonatank, line 88: aria-hidden is unnecessary, HDS icons default to `aria-hidden="true"` and you have to explicitly set it to false.

apps/admin-ui/src/component/ReservationListButton.tsx
```diff
@@ -17,43 +18,52 @@ export function ReservationListButton({
...
           {type === "deny" ? t("common.deny") : t("common.remove")}
         </Button>
       );
     case "restore":
       return (
-        <Button key={type} {...btnCommon} iconStart={<IconArrowUndo />}>
+        <Button
+          key={type}
+          {...btnCommon}
+          iconStart={<IconArrowUndo aria-hidden />}
```
joonatank, line 59: aria-hidden is unnecessary, HDS icons default to `aria-hidden="true"` and you have to explicitly set it to false.

joonatank (changes requested):

joonatank (approved):
The `aria-hidden` isn't a blocker, more cosmetic.

Merge to the monorepo: https://github.com/City-of-Helsinki/tilavarauspalvelu-core/pull/1534

